PR #1, opened 2015-10-21
STORY-2988 Fill in API Portal Documentation Gaps

codingiam:


source/api_documentation/transactions.rst
```diff
@@ -187,6 +187,66 @@ The data returned has the following fields:
...
+
+  * - dynamic_number_pool_referrer_param25
+    -
+    -
+
+  * - dynamic_number_pool_referrer_param25_name
+    -
+    -
+
+  * - dynamic_number_pool_referrer_param25_value
```
nburwell: please move these DNPR param ones down just below the existing param1 ~ 10 section (or, may need to go below the `Potentially Extra Params` section, but I think better would be to keep it with other params but just call out in the description that these 11 ~ 25 are only available as an add-on feature.

source/api_documentation/transactions.rst
```diff
@@ -187,6 +187,66 @@ The data returned has the following fields:
...
+
+  * - is_prepaid_data_append
+    -
+    -
+
+  * - do_not_call_data_append
+    -
+    -
+
+  * - gender
```
nburwell: Please move these down to the other `Data Append` ones (like age_range)

PR #9, opened 2015-11-27
91/t/STORY-3408/Documentation for Concurrent Call Cap Configuration via Network Integration

ibiziiac-sv:
Ticket: https://ringrevenue.atlassian.net/browse/STORY-3408
@leekirchhoff please review
cc: @dangrozasv

### DIFF
--- a/source/api_documentation/network_integration/advertiser_campaigns/index.rst
+++ b/source/api_documentation/network_integration/advertiser_campaigns/index.rst
@@ -9,7 +9,7 @@ Note that the `<advertiser_id_from_network>` and `<advertiser_campaign_id_from_n
 
 We support passing back current_terms and future_terms on campaigns. The current properties of the campaign are reflected in current_terms. All changes to the campaign are staged in future_terms. Once the campaign goes live, future_terms transition over to current_terms.
 
-You can set budgets on your campaign. There are two budget types, budget_cap_alert which is based on commissions, and call_cap_alert, which is based on the number of calls. These budget activities are only applicable for AffiliateEnabled campaigns (Known in the platform as a “Publisher Promotion” Campaign Type.)
+You can set budgets on your campaign. There are three budget types, budget_cap_alert which is based on commissions, call_cap_alert, which is based on the number of periodic calls, and concurrent_call_cap_alert, which is based on the number of simultaneous calls. These budget activities are only applicable for AffiliateEnabled campaigns (Known in the platform as a “Publisher Promotion” Campaign Type.)
 
 You are not allowed to delete campaigns.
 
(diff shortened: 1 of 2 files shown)


source/api_documentation/network_integration/advertiser_campaigns/index.rst
```diff
@@ -9,7 +9,7 @@ Note that the `<advertiser_id_from_network>` and `<advertiser_campaign_id_from_n
 
 We support passing back current_terms and future_terms on campaigns. The current properties of the campaign are reflected in current_terms. All changes to the campaign are staged in future_terms. Once the campaign goes live, future_terms transition over to current_terms.
 
-You can set budgets on your campaign. There are two budget types, budget_cap_alert which is based on commissions, and call_cap_alert, which is based on the number of calls. These budget activities are only applicable for AffiliateEnabled campaigns (Known in the platform as a “Publisher Promotion” Campaign Type.)
+You can set budgets on your campaign. There are three budget types, budget_cap_alert which is based on commissions, call_cap_alert, which is based on the number of periodic calls, and concurrent_call_cap_alert, which is based on the number of simultaneous calls. These budget activities are only applicable for AffiliateEnabled campaigns (Known in the platform as a “Publisher Promotion” Campaign Type.)
```
leekirchhoff: Should we be changing call_cap_alert to periodic_call_cap_alert, or would that be an incompatible change. Not sure how that name gets used.

Also instead of "which is based on the number of periodic calls" I recommend "which is based on the number of calls in a given period".

source/api_documentation/network_integration/advertiser_campaigns/index.rst
```diff
@@ -210,6 +210,22 @@ You are not allowed to delete campaigns.
...
     - Maximum Promo Numbers.
 
+  * - **concurrent_call_cap_alert**
+    -
+    -
+
+  * - starts_at
+    - date
+    - Call Cap Start. Default: Beginning of day with the campaign's time zone.
+
```
leekirchhoff: This is a little confusing for concurrent call caps, since they are continuous and don't really start at any particular time of day. Is this because we are following the pattern of other caps or can this be made more relevant to CCC?

source/api_documentation/network_integration/advertiser_campaigns/index.rst
ibiziiac-sv: We cannot rename `call_cap_alert` to `periodic_call_cap_alert ` at the API level because this change will affect all the versions of the API.

source/api_documentation/network_integration/advertiser_campaigns/index.rst
ibiziiac-sv: You are right, I have removed both `starts_at` and `time_zone` from the params list.

source/api_documentation/network_integration/advertiser_campaigns/index.rst
nburwell: With our DTOs, couldn't we bump a version and have the new name in new versions pretty easily?  Is that in the plan?

ibiziiac-sv: @nburwell I will write an email so we can discuss your idea with the team.

PR #11, opened 2015-12-09
Modification to files used for find / replace

spencerpbartz:
@mikeweaver 
These are the changes we made to the find / replace thing to make files that end in .tmp.


source/api_documentation/network_integration/networks/_create_users.rst
```diff
@@ -1,38 +0,0 @@
-.. container:: endpoint-long-description
```
spencerpbartz: @mikeweaver 
Did we have any kind of convention going for the collapsable GET / POST / PUT sections?
It looks like ```_[http_method]_[description].rst``` is the convention.

If that is the case, it looks like `network_integration/networks` had some unused files. There are 3 collapsable sections but 5 files. `_create_users.rst` is identical to `_post_network_users.rst` except for a newline.
I'd like to delete those to avoid confusing people.

source/conf.py
```diff
@@ -135,7 +135,7 @@ def build_partials(app, env, docnames):
       partial = open('{}{}{}'.format(source_path, docname, '.rst'), 'r').read()
```
spencerpbartz: @mikeweaver 
I couldn't figure out how to run a command in here like `find . -name "*.tmp" -type f -delete` at the end of the build. (maybe it belongs in the makefile anyhow).
I know they're not part of the repo but it might be nice to clean up the .tmp files when we're done. 
Are they of any use?

source/api_documentation/network_integration/networks/_create_users.rst
mikeweaver: I checked it out and agree, it looks unused. Kill it.

source/conf.py
mikeweaver: I wouldn't worry about them. I'm sure there is a way to do this in Python but it isn't occurring to me off the top of my head.

mikeweaver:
Looks good. Merge it!

PR #17, opened 2016-01-29
2016 01 25

oorion:
https://ringrevenue.atlassian.net/browse/STORY-3579

Adds new outcome columns to Transactions API.  Updates the transaction_type column to include the new "Outcome" type.


nburwell:
:+1: Changes look good (I don't know what the process is for the branch.. do we leave this branch as is? I assume we don't want to merge this into 2015-12-10 .. )

spencerpbartz:
@nburwell I followed up with @oorion about the procedure. 
There are some notes about how read-the-docs uses these branches at the beginning of the readme. (Don't know if the most recent changes I made to the current readme would be missing from new branches).

PR #20, opened 2016-03-02
STORY-3561 Support setting custom challenge prompt via API

codingiam:


spencerpbartz:
@codingiam: As per @mikeweaver 's request I checked the `doc_versions.py` file on this branch. Some of those dates have recently been updated, so `doc_versions.py` on this branch is now out of date. The dates from the most recent live version (2016-02-01) should be correct.

This should happen less frequently as time goes on, but there might be updates to the most recent API version **after** you have branched off of it. I'll add a note to the 2016-02-01 README, but we should probably notify the owners of any branches scheduled for **future** release so they can merge the current live version into their branch and carefully review any conflicts. 

I'm guessing what happened here was that @jebentier branched off of 2016-03-01 and hence got the outdated version of `doc_versions.py`

@jakesutton FYI ^

jebentier:
This all looks good.  We do not actually need to merge this branch into anything as it will be the new master once turned public on wednesday.

PR #21, opened 2016-03-07
2016 02 01 story 3745 dev portal small bugs

spencerpbartz:
@mikeweaver 
In light of recent RTD issues, this could use a once-over.

(Branch name is set to newest version but I can merge this in today.
I coordinated with Jake to ensure his new 2016-02-01 version gets these changes)


source/api_documentation/network_integration/advertiser_campaigns/_get_advertiser_campaign_unarchive.rst
```diff
@@ -0,0 +1,15 @@
+.. container:: endpoint-long-description
```
mikeweaver, line 1: @spencerpbartz This comment is probably not for you, but perhaps you can pass it on to the doc owner: it seems weird to support unarchiving on a GET command. GETs typically do not alter data on the remote side.

source/conf.py
```diff
@@ -409,7 +409,7 @@ def setup(app):
 .. raw:: html
 
   <link rel="stylesheet" href="https://media.readthedocs.org/css/sphinx_rtd_theme.css" type="text/css" />
-  <link rel="stylesheet" href="http://developers.invoca.net/en/""" + version+  """/_static/css/custom.css" type="text/css" />
+  <link rel="stylesheet" href="http://developers.invoca.net/en/""" + version +  """/_static/css/custom.css" type="text/css" />
```
mikeweaver: Do we have a ticket in the works to solve the protocol problem?

mikeweaver:
@spencerpbartz CR complete. Two questions, but they are non-blocking.

source/conf.py
nburwell: the way to do this (for all three of these resources here) is to remove the protocol entirely and the browser will load based on what the page is loaded as:

```
<link rel="stylesheet" href="//media.readthedocs.org/..."
<link rel="stylesheet" href="//developers.invoca.net/en..."
<link rel="stylesheet" href="//media.readthedocs.org/..."
```

source/api_documentation/network_integration/advertiser_campaigns/_get_advertiser_campaign_unarchive.rst
spencerpbartz, line 1: @mikeweaver 
Yeah it doesn't seem to follow convention if we're altering anything with GET. I think the problem is that we haven't been following convention for the history of the API.

@zmathe-sv 
Can you comment on this?

source/conf.py
spencerpbartz: @mikeweaver: Yes there is a ticket: [WEB-1960](https://ringrevenue.atlassian.net/browse/WEB-1960). Product has placed higher priority on [WEB-1958](https://ringrevenue.atlassian.net/browse/WEB-1958) which I am currently grooming / starting work on. 

@nburwell: Thanks for the advice! Mike also suggested that way might work, assuming we're going to stop trying to load the CSS file locally (which would require determining the path of the current working directory of each file at build time). I'll include you in the WEB-1960 discussion when we get there.

source/api_documentation/network_integration/advertiser_campaigns/_get_advertiser_campaign_unarchive.rst
zsmathe, line 1: Yes, I agree with you. I just replicated the examples above with go_live, where there's also a GET and a POST example. I think we should remove the GET example for the archive action. Would be ideal to remove the GET for go_live as well, but that's not possible without a version bump.
@mikeweaver, @spencerpbartz Do we also want to update the API endpoint to not allow GET request? This would be safe for the archive action since that wasn't exposed in the API doc till now.

mikeweaver, line 1: I would like to remove the GET action for archive from the API endpoint.

On Tue, Mar 8, 2016 at 4:41 AM, Zsolt Mathe notifications@github.com
wrote:

> In
> source/api_documentation/network_integration/advertiser_campaigns/_get_advertiser_campaign_unarchive.rst
> https://github.com/Invoca/developer-docs/pull/21#discussion_r55353596:
> 
> > @@ -0,0 +1,15 @@
> > +.. container:: endpoint-long-description
> 
> Yes, I agree with you. I just replicated the examples above with go_live,
> where there's also a GET and a POST example. I think we should remove the
> GET example for the archive action. Would be ideal to remove the GET for
> go_live as well, but that's not possible without a version bump.
> @mikeweaver https://github.com/mikeweaver, @spencerpbartz
> https://github.com/spencerpbartz Do we also want to update the API
> endpoint to not allow GET request? This would be safe for the archive
> action since that wasn't exposed in the API doc till now.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/Invoca/developer-docs/pull/21/files#r55353596.

## 

http://www.invoca.com/

_Mike Weaver_

Director of Software Engineering

m:| (805) 448-8900 <8054488900>

p: | (805) 880-7711 <8058807711>

skype: | mike.r.weaver

[image: Invoca Named to Inc. 500 List of Fastest-Growing Private Companies]
http://www.invoca.com/about/press/invoca-named-inc-500-list-fastest-growing-private-companies/
Invoca
Named to Inc. 500 List of Fastest-Growing Private Companies
http://www.invoca.com/about/press/invoca-named-inc-500-list-fastest-growing-private-companies/

spencerpbartz, line 1: @mikeweaver 
I looked through other parts of the NI API. 
Advertiser Campaigns seems to be the only place where GET requests do more than read.
I'll remove the GET **unarchive** example from this ticket. Shall I also rip out the existing GET **archive** and **go_live** examples?

spencerpbartz, line 1: @mikeweaver 
I discussed with @victorborda if it would be possible to rip out other non-RESTful API endpoints from the platform (and from the documentation).

There are a couple problems:
1. Customers might be using those as they stand today. If we deprecate them in the API we'll likely have a customer facing issue. 
2. Some customers may have old platforms and may not be able to use the correct HTTP verb, so we provide broader support while unfortunately breaking REST conventions. 

Given that, do we want to 
A) Remove the documentation but not the functionality? 
or 
B) Keep the documentation as it is and add this "unarchive" example?

mikeweaver:
As far as I know, the unarchive functionality is brand new, so no one
should be using it.

Unsure about customers that are unable to use HTTP verbs…

I think we have spent to much time talking about this already, it really
doesn’t matter in the grand scheme. Just ship the docs and code as-is.

On Thu, Mar 10, 2016 at 9:45 AM, spencerpbartz <notifications@github.com>
wrote:

> @mikeweaver <https://github.com/mikeweaver> What do you think of Victor's
> reasons?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Invoca/developer-docs/pull/21#issuecomment-194973256>.
>



-- 
<http://www.invoca.com/>

*Mike Weaver*

Director of Software Engineering

m:| (805) 448-8900 <8054488900>

p: | (805) 880-7711 <8058807711>

skype: | mike.r.weaver


[image: Invoca Named to Inc. 500 List of Fastest-Growing Private Companies]
<http://www.invoca.com/about/press/invoca-named-inc-500-list-fastest-growing-private-companies/>
Invoca
Named to Inc. 500 List of Fastest-Growing Private Companies
<http://www.invoca.com/about/press/invoca-named-inc-500-list-fastest-growing-private-companies/>

source/api_documentation/network_integration/advertiser_campaigns/_get_advertiser_campaign_unarchive.rst
spencerpbartz, line 1: As a result of a result of some disagreement in senior leadership we've decided to move forward as planned in the ticket

PR #23, opened 2016-03-10
update naming of signal to voice signal and call outcome to signal

oorion:
https://ringrevenue.atlassian.net/browse/STORY-3735

### DIFF
--- a/source/api_documentation/transactions_api/optional_fields.txt
+++ b/source/api_documentation/transactions_api/optional_fields.txt
@@ -42,43 +42,43 @@
 
   * - signal_1_value
     - Voice Signal: Credit Card
-    - Signals are configurable metrics that are tracked across calls, configured with rules about when the Signal is met (1 if Signal is met, 0 if Signal was applied to the call but not met, null if Signal was not applied to call)
+    - Voice Signals are configurable metrics that are tracked across calls, configured with rules about when the Voice Signal is met (1 if Voice Signal is met, 0 if Voice Signal was applied to the call but not met, null if Voice Signal was not applied to call)
 
   * - signal_2_value
     - Voice Signal: Competitors
-    - Signals are configurable metrics that are tracked across calls, configured with rules about when the Signal is met (1 if Signal is met, 0 if Signal was applied to the call but not met, null if Signal was not applied to call)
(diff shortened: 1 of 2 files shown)


jakesutton:
:+1:

spencerpbartz:
@oorion Nice clean up!
Looks good and is :+1: from me. I say this to everyone but my only advice is to pay reeeeeally close attention to those RST tables. From what I've gathered, both directives and RST tables need a full 2 new lines below them to work correctly buuut the Jury is still out.

source/api_documentation/transactions_api/optional_fields.txt
```diff
@@ -41,43 +41,43 @@
     - URL to the call recording, if available.
 
   * - signal_1_value
-    - Signal: Credit Card
+    - Voice Signal: Credit Card
     - Signals are configurable metrics that are tracked across calls, configured with rules about when the Signal is met (1 if Signal is met, 0 if Signal was applied to the call but not met, null if Signal was not applied to call)
```
nburwell: Should we say "Voice signals are configurable ..... and when the Voice signal is met (1 if Voice signal is met) ..." etc?  To drive home that the 1, 0, null is for voice signals only.   

Since further below our `signal_value` (which is for "outcomes") is true/false as its value.

source/api_documentation/transactions_api/optional_fields.txt
```diff
@@ -387,32 +387,32 @@
...
+    - Signal Occurred At
+    - 10 digit time that the signal occurred, in UTC seconds since 1/1/70, also known as Unix time_t.
+
+  * - signal_source
+    - Signal Source
+    - The source of the signal.  Possible values are :UserOverride, :Api, :Import, :Expression, :Ivr, and :Machine
+
+  * - signal_value
+    - Signal Value
+    - True or false as to whether the signal was met or not.
```
nburwell: BTW is this field null for transactions that do note have a call_outcome attached to it? If so we should improve the docs with that note (similar to what we say in voice signal columns)

source/api_documentation/transactions_api/optional_fields.txt
jakesutton: Agree :100:

PR #24, opened 2016-03-10
update naming of signal to voice signal and call outcome to signal

oorion:
https://ringrevenue.atlassian.net/browse/STORY-3735


jakesutton:
:+1:

PR #31, opened 2016-03-25
2016 02 01 web 1958 api dev portal transaction api parameter list fixes

spencerpbartz:
@mikeweaver @nburwell @jebentier 
Here are the changes to make proper transactions API parameter lists for the developer docs portal. 
Code Review requested.

FYI: @codingiam @zmathe-sv 
The 2016-03-01 branch is not live yet so we will need to back merge these changes before we make it the new "latest version."


source/api_documentation/transactions_api/_advertiser_optional_param_table.rst
```diff
@@ -0,0 +1,423 @@
...
+    - Call Outcome Value
+    - Whether the outcome was met, 1 = outcome met, 0 = outcome not met (defaults to 1).
+
+  * - cell_phone_number
+    - Cell Phone (Reported)
+    - [Call center integration only] Customer's cell phone number as given to the call center.
+
+  * - country
+    - Country (Reported)
+    - [Call center integration only] Customer's country as given to the call center.
```
nburwell: I feel like we should keep the "(Reported)" fields (aka those from Call center integration / CPA enabled) together.  Having them interspersed is confusing I think.

In general I think grouping the optional / feature-based columns by feature, will make more sense, versus alphabetical.

source/api_documentation/transactions_api/_advertiser_optional_param_table.rst
```diff
@@ -0,0 +1,423 @@
...
+
+  * - dynamic_number_pool_referrer_ad_id
+    - AdWords Ad ID
+    - AdWords Ad ID, provided by Google
+
+  * - dynamic_number_pool_referrer_param1
+    - Dynamic Number Pool Referrer Param1 Value
+    - [Number pooling links only] name=value for parameter 1, url escaped. Example: name%3Dvalue
+
+  * - dynamic_number_pool_referrer_param10
```
nburwell: similar theme here.. I don't think strict alphabetical makes sense here.. I would show param1, param2, ... param10 then the 11 ~ 25 in numeric order after that (not sure if we need a header for the 11~25 params or not?)

source/api_documentation/transactions_api/_advertiser_param_table.rst
```diff
@@ -0,0 +1,239 @@
...
+
+  * - region
+    - Region
+    - Region (state, province or country) where transaction originated
+
+  * - repeat_calling_phone_number
+    - Repeat Caller
+    - Whether the call was a repeat call. Repeat call detection is not applied to shared or unavailable caller ids.
+
+  * - signal_10_value
```
nburwell: here too.. let's put signal_10 at the end, even though it won't be strict alphabetical

source/api_documentation/transactions_api/_affiliate_optional_param_table.rst
```diff
@@ -0,0 +1,387 @@
+
+..  list-table::
+  :widths: 30 8 40
+  :header-rows: 1
+  :class: parameters
+
```
nburwell: all my comments from advertiser optional param table apply here =)

source/api_documentation/transactions_api/_network_optional_param_table.rst
```diff
@@ -0,0 +1,411 @@
+
+..  list-table::
+  :widths: 30 8 40
+  :header-rows: 1
+  :class: parameters
```
nburwell: same comments here regarding order and grouping

source/api_documentation/transactions_api/_network_optional_param_table.rst
```diff
@@ -0,0 +1,411 @@
...
+    - Keywords
+    - [Number pooling links only] Search keywords used
+
+  * - dynamic_number_pool_referrer_search_keywords_id
+    - AdWords Keywords ID
+    - AdWords Keyword ID, provided by Google
+
+  * - dynamic_number_pool_referrer_search_type
+    - Search Type
+    - [Number pooling links only] "Paid" or "Organic".
```
nburwell: can you do `include`'s from within these rst files? If so I would consider making the 30+ dynamic_nuber_pool\* fields into its own rst file that each of the three could pull in. Same with the CPA columns and others that is a group of 10+ that show up in all reports. That way the lists of optional params per org type would hopefully read pretty short in the source

nburwell:
General question @spencerpbartz :

Since this is a bug fix for missing fields, should we also move this into all live branches on the API portal? I think we should so that we are accurate in what we have in the docs (we will just have to be careful to omit some params for those older branches, but that info should be right here: https://github.com/Invoca/web/blob/master/app/controllers/api/transactions_api_controller.rb#L120-L126 )

nburwell:
And nice work on the new files, I'm happy with the set of fields going in per org type! 

This PR is showing pretty much all green (new stuff), where is the merge / delete of the old ways of presenting all the transaction API params?

spencerpbartz:
@nburwell Thanks for the review!
General question: It is certainly possible to back-merge this into old versions, and we should for accuracy. It will be a bit more work since older versions have not been rebuilt since RTD changed CSS include order and also started injecting ads.  2016-01-25 should be easy to fix but the previous version might be tricky as it didn't didn't even have separate sections for org_type. 

Green / New stuff: 
Oops. Yeah the old param table is D.O.A. at this point. 
Like in Poltergeist I removed the gravestones but not the bodies. 
I'll `git rm` the old param list files.

source/api_documentation/transactions_api/_network_optional_param_table.rst
spencerpbartz: @nburwell According to the docs we can't, but the template file use to build all 3 org_types can!
(`source/custom_templates/_tx_api_page.txt`)

source/custom_templates/_tx_api_page.txt
```diff
@@ -1,8 +1,9 @@
 .. include:: _:page:.tmp
 
-.. include:: common_response_params.rst
+.. include:: _:org_type:_param_table.rst
 
-.. include:: optional_fields.txt
+The following fields are optional based on your account type:
 
+.. include:: _:org_type:_optional_param_table.rst
```
spencerpbartz: @nburwell Right here we could include DNPR and CPA stuff

source/api_documentation/network_integration/affiliates/_put_affiliate.rst
```diff
@@ -6,7 +6,7 @@
 
   Endpoint:
```
spencerpbartz, line 8: In case anyone wonders, this is unrelated to the story but was a typo Bobby found so I threw it in.

source/custom_templates/_tx_api_page.txt
nburwell: Cool, if we can make all the optional fields still look coherent and flow nicely in the final doc output page, I think it's worth it to reduce duplication in a lot of spots.

source/api_documentation/transactions_api/_advertiser_optional_param_table.rst
spencerpbartz: @nburwell Great point. I hadn't been paying attention to the finer details of each param. 
Putting them in a separate tables would be the easiest thing, especially if they are the same for all org_types.

source/custom_templates/_tx_api_page.txt
spencerpbartz: @nburwell It doesn't look like I can build an RST table with rows stored in different files. Is it OK if they are separate tables?

nburwell: I think so, would be nice if we had an overall header for the section, and then each table could be annotated in some way:

---

## Optional Parameters

#### Signal Parameters

[table]

#### Conversion Reporting Parameters

[table about call center integration fields]

#### Additional Features

[table with the rest]

Please collaborate with Nathan and get his input on the headers / grouping, but I think something like the above should work.

spencerpbartz: @nburwell Do I have these right? 
1. Call center integration params are included in Network reports by default (and are therefore not optional to Networks)? 
2. Signal params are never available to Affiliates and sort of optional for Network and Advertiser? (Looking at the DetailReport code though it seems that the majority of customers would get signal data out of the box, our dummy users in the scripts got Signal params from static columns not dynamic).

If the above is correct, it's going to be a little more work to try and make this DRY since all 3 org_types use the same template. I can make it work with a little massaging though.

nburwell: No, that is not correct.

1) Call center integration params are optional (only if `cpa_enabled?`: https://github.com/Invoca/web/blob/master/app/models/reports/detail/detail_report.rb#L679-L689)

2) Signal params are optional (and only for Network and Advertiser)
https://github.com/Invoca/web/blob/master/app/models/reports/detail/detail_report.rb#L564-L567 (ends up checking a feature flag)

So what I had in my above comment should be accurate (in that these two tables would go in optional sections). It looks like neither are shown for Affiliates though, hopefully that doesn't add much work..

spencerpbartz: Thanks for the clarification. 
So 1) Is not a problem but 2) is because one org_type needs to be catered to specially. 
I'll find a way!

spencerpbartz:
@nburwell 
CR comments implemented. Make sure you're logged in to RTD and
[click here to see the private branch:](https://invoca-developer-docs.readthedocs.org/en/2016-02-01_web-1958_api_dev_portal_transaction_api_parameter_list_fixes/api_documentation/transactions_api/index.html)

nburwell:
Some general comments (some of these will be improvements on top of what you've already done, but easy wins while we are here):
- Can you decrease the font-size of each section, we need a sub-heading style..
  image
- Please rename the "Dynamic Number Pool Referrer Parameters" section to `RingPool Parameters` (that's the user-facing / platform-facing name)
- Let's rename the "Name in Reports" column of the rows that have "Dynamic Number Pool Referrer ParamX Value" to be just "Pool Param X" (hint: these never show up in UI reports, so that name being used here was an ugly generated one.. shortening it will prevent wrapping and make it match the others)
- Can we remove `[Number pooling links only]` from the description field of all these RingPool params that have it? We don't refer to the feature as "Number pooling" ever, and it's not for links only, it's for API access and more, so that disclaimer is just inaccurate and confusing.
- Please remove all smart quotes
  image
- Please group all the AdWords columns together (in the RingPool Parameters table)
- Please rename "Call Center Integration Parameters" section to be "Conversion Reporting Parameters"
- I would make "Data Append" columns (that have "Data from caller demographics lookup" in description) all in their own table too -- the section title should be "Enhanced Caller Profile Parameters"
- Let's reword "The following fields are optional based on your account type:" to something like "The following fields are optional based on your account type or enabled features:"

If you have questions on any of these, let's work through them together in person.

source/api_documentation/transactions_api/_affiliate_optional_param_table.rst
```diff
@@ -0,0 +1,41 @@
...
+  * - bundled_destination_ids
+    - Bundled Destination Ids
+    - 
+
+  * - bundled_destination_ids_from_network
+    - Bundled Destination Ids From Network
+    - 
+
+  * - call_outcome_outcome_name_2
```
nburwell: this should be `call_outcome_outcome_name` not `call_outcome_outcome_name_2`

source/api_documentation/transactions_api/_affiliate_optional_param_table.rst
```diff
@@ -0,0 +1,41 @@
...
+  :header-rows: 1
+  :class: parameters
+
+  * - Field
+    - Name in Reports
+    - Description
+
+  * - bundled_destination_ids
+    - Bundled Destination Ids
+    -
```
nburwell: Where is this description? Let's add one (likely there was in the dev portal, and maybe help text was missing it?)

source/api_documentation/transactions_api/_affiliate_optional_param_table.rst
```diff
@@ -0,0 +1,41 @@
...
+    - Name in Reports
+    - Description
+
+  * - bundled_destination_ids
+    - Bundled Destination Ids
+    - 
+
+  * - bundled_destination_ids_from_network
+    - Bundled Destination Ids From Network
+    -
```
nburwell: Let's add a description here

source/api_documentation/transactions_api/_affiliate_optional_param_table.rst
```diff
@@ -0,0 +1,41 @@
...
+    - Bundled Destination Ids
+    - 
+
+  * - bundled_destination_ids_from_network
+    - Bundled Destination Ids From Network
+    - 
+
+  * - call_outcome_outcome_name_2
+    - Outcome Name
+    -
```
nburwell: Let's add the outcome name description here [whatever the network side said probably]

source/api_documentation/transactions_api/_advertiser_optional_param_table.rst
```diff
@@ -0,0 +1,77 @@
...
+
+  * - bundled_destination_ids
+    - Bundled Destination Ids
+    - 
+
+  * - bundled_destination_ids_from_network
+    - Bundled Destination Ids From Network
+    - 
+
+  * - call_outcome_custom_parameter_1
```
nburwell: these 3 custom params should have descriptions (they did in some PR I saw come through from Jake to add these to the dev portal..)

nburwell:
The point of requesting Signal, Conv Reporting, Pool params etc being in their own tables, was to prevent so much duplication.. but there is a `_network_signal_param_table` and `_advertiser_signal_param_table` file that are the same?  I think I know how you got there, but there must be a different way and I think it's re-thinking / re-doing how `custom_templates/_tx_api_page.txt` works. Let's connect tomorrow!

source/api_documentation/transactions_api/_advertiser_optional_param_table.rst
spencerpbartz: OK. I can dig those up. 
But that leads me to ask why those descriptions don't show up with the technique we used? 
Shouldn't they be available?

spencerpbartz:
@nburwell 
Thanks for the UX and architectural perspective on this.  
The specific requirements of each organization has made avoiding duplication pretty difficult. The only params common to all 3 organizations are DNPR.

We don't have the luxury of any type of server side logic after the build is complete so the options I can think of are:
1. Build all files the same, then use JavaScript and / or CSS after the fact to hide certain areas. 
2. Modify the build script to treat certain files differently. (if org_type == affiliate, don't include certain files).
3. Have a _net_adv_tx_api_page.txt template because those org_types seem to share more with each other than with affiliates, then have a separate _aff_adv_tx_api_page.txt template. 

Keeping in mind there is no more time left on this ticket, I would vote for #1 as it would result in the smallest number of files / changes. I have a feeling @mikeweaver won't like any of the above so I'm open to other ideas. 
1. Additionally we could revert the 3 orgs in 1 system and not use the _tx_api_page.txt template at all.

spencerpbartz:
@nburwell "Pool Param 1" still wraps for me. 
How about just getting rid of the column since I'm basically just making up names?

nburwell:
Pool Param 1 works for me.. I hope for 90+% of the rows you didn't make up names but got them via our script we iterated on, that get's the non-API friendly name, right?

spencerpbartz:
Thanks for bearing with me, I haven't worked with RingPools so I just know the high level. 
So yes, any names I had came from the scripts and I'm just shortening them.
I was wondering mostly about the sets of 3: 
`dynamic_number_pool_referrer_param1`
`dynamic_number_pool_referrer_param1_name`
`dynamic_number_pool_referrer_param1_value`

Which had the following "Name in reports" entries respectively:
`Dynamic Number Pool Referrer Param1 Value`
`Pool Param 1 Name`
`Pool Param 1 Value`
1. Should I just change the first one to `Pool Param 1` and leave the others as-is? 
2. What should I call the column header if these don't show up in reports? Maybe just "name" ?

spencerpbartz:
@nburwell @mikeweaver 
Round 2 of CR implementation complete.
View it [here](https://invoca-developer-docs.readthedocs.org/en/2016-02-01_web-1958_api_dev_portal_transaction_api_parameter_list_fixes/api_documentation/transactions_api/index.html).

source/api_documentation/transactions_api/advertiser_user.rst
```diff
@@ -1,7 +1,71 @@
...
+
+.. code-block:: bash
+
+  curl -k "https://mynetwork.invoca.net/advertisers/transactions/33.csv?limit=20&start_after_transaction_id=C624DA2C-CF3367C3&oauth_token=YbcFHZ38FNfptfZMB0RZ6dk9dOJCaCfU"
+
+The -k option asks curl to not bother checking the SSL certificate authority chain as that requires extra configuration.
+
+Endpoint:
+
+``https://invoca.net/api/@@TRANSACTION_API_VERSION/advertisers/transactions/``
```
nburwell, line 75: nice, I think this file structure is much easier to follow and will pay off in the balance of sharing common sets of params in one place, but having flexibility per org type for one-off differences

source/api_documentation/transactions_api/advertiser_user.rst
```diff
@@ -1,7 +1,71 @@
...
+The data returned has the following fields:
+
+.. include:: advertiser_param_table.rst
+
+Optional Parameters
+-------------------
+
+The following fields are optional based on your account type:
+
+.. include:: dnpr_param_table.rst
```
nburwell, line 50: This advertiser page is missing the `adwords_param_table.rst`

source/api_documentation/transactions_api/affiliate_user.rst
```diff
@@ -1,7 +1,69 @@
...
+then repeat as necessary to get all call records within date range.
+
+Response
+--------
+
+The data returned has the following fields:
+
+.. include:: affiliate_param_table.rst
+
+.. include:: conversion_reporting_param_table.rst
```
nburwell: I confirmed in the last round that affiliates do not have conversion reporting.. so this line can be removed.  

Please also update the script in tools repo to not override `cpa_enabled?` on affiliate report class (it's impossible to be true based on the actual repo code) https://github.com/Invoca/web/blob/master/app/models/reports/detail/detail_report.rb#L857-L857

source/api_documentation/transactions_api/network_user.rst
```diff
@@ -1,7 +1,75 @@
...
+.. list-table::
+  :widths: 8 40
+  :header-rows: 1
+  :class: parameters
+
+  * - Format
+    - Description and URL
+
+  * - csv
+    - Comma-Separated Values, or really Anything-Separated Values (see column_separator= below). Returns an optional header row followed by one row for each transaction, with delimited values for each row. ``https://mynetwork.invoca.net/api/@@TRANSACTION_API_VERSION/advertisers/transactions/33.csv``
```
nburwell: this url should have `/networks/` not advertisers

nburwell:
2nd round of review done! 

I know that last refactor was a pain but I'd say definitely success when so much of the special case config / python code goes away!  And we now have more stuff shared, and I like that it's more grouped by features now.

> 1) Should I just change the first one to Pool Param 1 and leave the others as-is?

Yes, What you did is great. 

> 2) What should I call the column header if these don't show up in reports? Maybe just "name" ?

Please match the header with all the others "Name in Reports".  I only meant that the special case one (that you mapped to Pool Param 1) does not show in reports technically. But "Pool Param 1 Name" and "Pool Param 1 Value" _do_ show up in reports, 1 ~ 25 do.  So overall for the table it's still appropriate to have that heading.

PR #36, opened 2016-04-07
2016-02-01 Update json to appear red

oorion:


jakesutton:
:+1:

spencerpbartz:
@oorion Looks good to me. Please merge into **2016-02-01** and **2016-03-01** when you're ready. 
It looks like `_create_signal2.rst` and `_update_signal2.rst` were added in **2016-02-01** so you might try cherry picking everything except those 2 when you merge into **2016-01-25** and **2015-12-10.**

I ran into a similar situation with the Transactions API params so let me know if you need help.

PR #38, opened 2016-04-11
96 u documentation updates for STORY-3828

eric-dowty:
This PR adds information to the SAPI documentation for supported timestamp formats. 

Below is a picture of the descriptions:

screen shot 2016-04-11 at 4 42 05 pm

### DIFF
--- a/source/api_documentation/signal_api/index.rst
+++ b/source/api_documentation/signal_api/index.rst
@@ -143,6 +143,7 @@ Endpoint:
   :description: Update a signal
   :page: update_signal2
 
+
 Timestamp Formats
 -------------------------------
 
@@ -182,6 +183,9 @@ All examples below correspond to a date time of **11 April 2016** at **1 PM Paci
     Example: **2016/04/11 13:00:00.000 PM**
 
     Example (no milliseconds): **2016/04/11 13:00:00 PM**
+
+    **Note:** The file upload user interface includes a timezone dropdown feature, which allows users to select which timezone this format is parsed in.
+
 
 
 Example POST Request Using cURL
--- a/source/api_documentation/signal_api/index.rst
+++ b/source/api_documentation/signal_api/index.rst
@@ -96,11 +96,11 @@ Used to create the fields of a signal.
 
     `value:` True or false as to whether the signal was met or not. Defaults to true if not passed. Can be a string ‘true’ or ‘false’, or 1 (true) or 0 (false).
 
-    `custom_paramter_1` Custom parameter. Up to 255 character string.
+    `custom_parameter_1` Custom parameter. Up to 255 character string.
 
-    `custom_paramter_2` Custom parameter. Up to 255 character string.
+    `custom_parameter_2` Custom parameter. Up to 255 character string.
 
-    `custom_paramter_3` Custom parameter. Up to 255 character string.
+    `custom_parameter_3` Custom parameter. Up to 255 character string.
 
 ------
 
@@ -173,15 +173,15 @@ All examples below correspond to a date time of **11 April 2016** at **1 PM Paci
 
     Example (UTC): **2016/04/11T20:00:00.000Z**
 
-    Example (without milliseconds): **2016/04/11T13:00:00-07:00**
+    Example (no milliseconds): **2016/04/11T13:00:00-07:00**
 
 **Excel Compatible:** Timestamp parsed in the timezone of the **oauth token**'s associated network. Milliseconds are optional.
 
     Format: **YYYY/MM/DD HH:MM:SS.sss AM/PM**
 
-    Example (with milliseconds): **2016/04/11 13:00:00.000 PM**
+    Example: **2016/04/11 13:00:00.000 PM**
 
-    Example (without milliseconds): **2016/04/11 13:00:00 PM**
+    Example (no milliseconds): **2016/04/11 13:00:00 PM**
 
 
 Example POST Request Using cURL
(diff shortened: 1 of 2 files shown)


source/api_documentation/signal_api/index.rst
```diff
@@ -144,6 +143,42 @@ Endpoint:
...
+
+**ISO 8601:** Timestamp with +/- UTC offset or Z to indicate time is in UTC.
+
+    Format: **YYYY/MM/DDTHH:MM:SS.sss+hh:mm**
+
+    Example (UTC offset of +3 hours): **2016/04/11T23:00:00.000+03:00**
+
+    Example (UTC offset of -7 hours): **2016/04/11T13:00:00.000-07:00**
+
+    Example (UTC): **2016/04/11T20:00:00.000Z**
```
nburwell, line 175: can you please show an example without milliseconds? 

(We support both of those on the code side, right..? ) https://www.w3.org/TR/NOTE-datetime

source/api_documentation/signal_api/index.rst
```diff
@@ -144,6 +143,42 @@ Endpoint:
...
+
+    Format: **YYYYMMDDHHMMSSsss**
+
+    Example: **20160411130000000**
+
+**Standard:** Timestamp parsed in the timezone of the **oauth token**'s associated network.
+
+    Format: **YYYY/MM/DD HH:MM:SS.sss AM/PM**
+
+    Example: **2016/04/11 13:00:00.000 PM**
```
nburwell, line 183: ditto here I hope the `.sss` is optional, and we should say that and show an example without it

source/api_documentation/signal_api/index.rst
```diff
@@ -144,6 +143,42 @@ Endpoint:
...
+
+    Example (UTC): **2016/04/11T20:00:00.000Z**
+
+**Compressed:** 17 digit timestamp always parsed in Pacific time.
+
+    Format: **YYYYMMDDHHMMSSsss**
+
+    Example: **20160411130000000**
+
+**Standard:** Timestamp parsed in the timezone of the **oauth token**'s associated network.
```
nburwell: as discussed, "Standard" sounds like it's what we want the users to use... or that it's like an ISO standard.. neither of which I think is true =)

Since this is the format most friendly for / most used by M$ Excel, what about `**Excel Compatible:**` ?

source/api_documentation/signal_api/index.rst
eric-dowty, line 183: Will do. sss is definitely not needed, so an additional example here (and for the comment above) is warranted.

nburwell:
Not related to your change, but something Nathan noticed on the portal on this section, and since you already have a PR and the doc tool working.. can you fix the typos on this page around custom param names?
https://github.com/Invoca/developer-docs/blob/2016-03-01_update_time_zone_for_sapi/source/api_documentation/signal_api/index.rst

```
custom_paramter_1 => custom_parameter_1
custom_paramter_2 => custom_parameter_2
custom_paramter_3 => custom_parameter_3
```

nburwell:
And one last favor from Nathan, pleeease =) to remove "Telecom" from the "Telecom Competitors" phrase on this page, so signal_8_value title should just be "Voice Signal: Competitors":

https://github.com/Invoca/developer-docs/blob/2016-03-01_update_time_zone_for_sapi/source/api_documentation/transactions_api/voice_signal_param_table.rst

You rock!

source/api_documentation/transactions_api/voice_signal_param_table.rst
```diff
@@ -39,7 +39,7 @@ Voice Signal Parameters
     - Voice Signals are configurable metrics that are tracked across calls, configured with rules about when the Voice Signal is met (1 if Voice Signal is met, 0 if Voice Signal was applied to the call but not met, null if Voice Signal was not applied to call)
 
   * - signal_8_value
-    - Voice Signal: Telecom Competitors
+    - Voice Signal: Competitors
     - Voice Signals are configurable metrics that are tracked across calls, configured with rules about when the Voice Signal is met (1 if Voice Signal is met, 0 if Voice Signal was applied to the call but not met, null if Voice Signal was not applied to call)
```
eric-dowty, line 43: @nburwell Does this need to be updated for every API version? Or just the 2016-03-01 version?

source/api_documentation/signal_api/index.rst
```diff
@@ -144,6 +143,46 @@ Endpoint:
...
+
+    Example (UTC offset of +3 hours): **2016/04/11T23:00:00.000+03:00**
+
+    Example (UTC offset of -7 hours): **2016/04/11T13:00:00.000-07:00**
+
+    Example (UTC): **2016/04/11T20:00:00.000Z**
+
+    Example (no milliseconds): **2016/04/11T13:00:00-07:00**
+
+**Excel Compatible:** Timestamp parsed in the timezone of the **oauth token**'s associated network. Milliseconds are optional.
```
nburwell, line 179: Do we want to say here that when using this format in File Upload case, that the timezone can be overridden per upload? 

Since otherwise I could see a user reading this after being linked to it from the File Upload uservoice article and thinking they are stuck

source/api_documentation/transactions_api/voice_signal_param_table.rst
nburwell, line 43: Nathan is fine with this just being on this newest version

spencerpbartz:
@eric-dowty You're good to merge :+1:

PR #43, opened 2016-04-12
2016 01 25 update time zone for sapi

eric-dowty:
Updates to timestamp formats in conjunction with STORY-3828:

https://ringrevenue.atlassian.net/browse/STORY-3828

Screen shot of primary change:

5d1bd0ca-0004-11e6-941a-aa87d375a137

### DIFF
--- a/source/api_documentation/signal_api/index.rst
+++ b/source/api_documentation/signal_api/index.rst
@@ -49,7 +49,7 @@ Used to find the call associated with the signal.
 
     or
 
-    `start_time_t:` 10 digit start time of the call in UTC seconds since 1/1/70, also known as Unix time_t. UTC milliseconds since 1/1/70 (which is the default in Javascript) are also supported, i.e. a 13 digit start time. The start time does not have to be exact since clocks will not be perfectly synchronized. We will find the closest match (in combination with **duration_in_seconds**, if provided)
+    `start_time_t:` The start time does not have to be exact since clocks will not be perfectly synchronized. We will find the closest match (in combination with **duration_in_seconds**, if provided). See **Timestamp Formats** section below for descriptions of supported timestamps.
 
     Note: If using **start_time_t** to match, instead of **transaction_id**, one or more of the following optional parameters is highly recommended. The optional parameters are ignored when using the **transaction_id**:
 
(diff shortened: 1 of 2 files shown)


nburwell:
:+1:

spencerpbartz:
The display looks fine but there are still warnings on this branch's build:
The build script is an eruption of cacophonous incomprehensibility, but I built it locally and noticed this warning
`/Users/sbartz/projects/developer-docs/source/api_documentation/signal_api/index.rst:230: WARNING: Explicit markup ends without a blank line; unexpected unindent.`

Checking that line leaves us with no clues so I looked upward to where our custom directives are being used:
image

The warning goes away if we add another newline in between.

```
:page: update_signal2

Timestamp Formats
-------------------------------
```

`\n` in RST does not map to `<br>` in this case so you are safe adding a few new lines. 
A similar situation happens with table directives and can cause all or part of a table not to show up. 
The line number in the error message is pretty useless as it leads you in the wrong direction. 

@mikeweaver This warning was probably missed because of the flood of `WARNING: document is not included in any toctree` and also our own `BUILDING PARTIAL:` messages. 
I dug into the sphinx source commented out the toctree stuff on my local **venv** **here:** `venv/lib/python2.7/site-packages/sphinx/environment.py:2012:`
We don't want those documents to be in toctrees in the first place because they're partials. 
We might want to think about our build script output to prevent future confusion.

@eric-dowty Good work on this one. As you get more used to RTD you'll get used to how warnings manifest themselves in the build output and be mindful of indentation and new lines. So just fix that warning then you're :+1:

mikeweaver:
@spencerpbartz Thanks for looking at this. A few notes:

The  api_endpoint macro was made by us. Its possible that after it expands in RST it isn't adding enough trailing white space. Can you look at that? If that is the cause, it should be easy to fix.

Regarding the flood of warnings: now that you have found out where it is, we should be able to monkey patch over it. Lets talk tomorrow about how we can do this.

spencerpbartz:
I did some investigating. Interesting... An echo of Python perhaps. 
It's not so much the line spacing but the indentation.

In general, directives do not have a  "begin / end." They just begin and then continue until they 
 either 1. Run into a new line then text or whitespace,  2. Run into another directive whose indentation is <= to theirs. I reproduced this with a built-in directive So ours is just following suit. I can explain more tomorrow. 
The code below is _syntactically correct_
image

**The code below is not**
image

spencerpbartz:
:+1:

PR #55, opened 2016-04-26
WEB-2101 remove incorrect statement

liamcardenas:
@jebentier @Inanepenguin @aschreifels 

cc: @mikeweaver

### DIFF
--- a/source/api_documentation/transactions_api/intro.rst
+++ b/source/api_documentation/transactions_api/intro.rst
@@ -4,7 +4,7 @@ Authentication
 --------------
 
 The API uses OAuth Authentication to authenticate that access is allowed. Pass the OAuth Token like any other query parameter, however, please note that the OAuth token is a required parameter.
-OAuth Tokens may be generated from the Manage API Credentials page. You must be a super user to get an OAuth Token and use the API.
+OAuth Tokens may be generated from the Manage API Credentials page.
 
 Query Parameters
 ----------------
@@ -51,5 +51,3 @@ Please note, the "to" and "from" date range parameters are both necessary, provi
 Example:
 
 Send request 1:
-
-
--- a/source/api_documentation/transactions_api/intro.rst
+++ b/source/api_documentation/transactions_api/intro.rst
@@ -51,3 +51,5 @@ Please note, the "to" and "from" date range parameters are both necessary, provi
 Example:
 
 Send request 1:
+
+


source/api_documentation/transactions_api/intro.rst
```diff
@@ -51,5 +51,3 @@ Please note, the "to" and "from" date range parameters are both necessary, provi
 Example:
 
 Send request 1:
-
-
```
peterwilson, line 55: Were these lines needed for spacing? Just making sure your IDE didn't remove these automatically/unintentionally.

jebentier:
Things look good to me aside from the comment @Inanepenguin raised.

liamcardenas:
@Inanepenguin @jebentier 
Oops! I should have checked this before. That was definitely not intentional, I'm not sure why it did that. It is weird that it did this on each branch even though I didn't cherrypick the commit. I don't think it was the editor because I was just using atom.

Anyway, I'll look into it, thanks for pointing this out!

liamcardenas:
@Inanepenguin 
Ok, I looked into it and I have no idea why it says that was deleted. It was never there to begin with, for example: https://github.com/Invoca/developer-docs/blob/2016-03-01/source/api_documentation/transactions_api/intro.rst (look at the very bottom)

Also, when I add it in there and re-build it, it does not provide any relevant styling. I'm not sure how those dashes got on my branches, but they certainly should be deleted and were never meant to be there.

liamcardenas:
@Inanepenguin should I merge?

peterwilson:
@liamcardenas compare yours: https://github.com/Invoca/developer-docs/blame/98/eb/WEB-2101_remove_false_permissions_claim_about_oauth_token_2016-01-25/source/api_documentation/transactions_api/intro.rst and the branch you're merging into: https://github.com/Invoca/developer-docs/blame/2016-01-25/source/api_documentation/transactions_api/intro.rst

Those two lines were definitely there. I'm concerned that they are intentional for spacing. Most viewers/editors will remove white space at the end of files, but this file isn't going to be displayed individually. It is aggregated with other files, so it won't be whitespace at the end of a file. Can you add the two lines back before merging?

liamcardenas:
@Inanepenguin 
Ah I misread github and thought the lines had dashes on them (but those were just minus signs)

Regardless, the whitespace at the end of the file makes no difference (it does not even affect spacing). I have tested it. Do you still want me to add it back in?

peterwilson:
@liamcardenas as silly as it may seem, it might be a wise choice to add them back in. We've seen some insanely subtle bugs because of whitespace or weird new lines. Maybe @mikeweaver can weigh in here.

Thanks for double checking that it's not needed. Maybe it was updates in a recent version, or maybe those lines were truly unnecessary, but I'd be concerned we missed something.

mikeweaver:
@liamcardenas Read the docs is an incredibly finicky system. If we don't have to remove the whitespace, then don't remove it. For RTD, its "better safe than sorry".

liamcardenas:
@Inanepenguin fixed!
cc: @mikeweaver

peterwilson:
:+1:

PR #62, opened 2016-05-13
98/eb/web 2123 remove orphan tag

liamcardenas:
@mikeweaver 

Here is the first one. I opted for creating a bash script, so the new build process is:
"bash html" instead of "make clean && make html"

I'm sure this could've been done with Make, but it is not nearly as straightforward. Let me know your thoughts so I can move forward on the older branches. Thanks!

@spencerpbartz: you might find this interesting


html
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+make clean
+make html 2> >(grep -v "WARNING: document isn't included in any toctree" 1>&2)
```
mikeweaver: We only want to exclude warnings for files that start with a _

src/pip-delete-this-directory.txt
```diff
@@ -0,0 +1,5 @@
+This file is placed here by pip to indicate the source was put
```
mikeweaver: These two files should not have been committed. Always check your:
- git status before you commit
- Your PR before you send it out for review

html
liamcardenas: Ah, alright!

liamcardenas:
@mikeweaver ok I fixed those things, I am going to move ahead on the other branches. Let me know if you have any other feedback

.gitignore
```diff
@@ -6,3 +6,4 @@ html/
 *.pyc
 *.swp
 *.tmp
+src/
```
mikeweaver, line 9: Ignore the src folder? What folder is that?

liamcardenas, line 9: This is a folder than is sometimes automatically generated by pip when you run:
pip install -r requirements.pip

We don't have any of our own code in this folder.

README.md
```diff
@@ -24,7 +24,7 @@ sudo pip install virtualenv
 virtualenv venv
 source venv/bin/activate
 pip install -r requirements.pip
-make html
+bash html
```
spencerpbartz, line 27: What if I want to use my secret target that removes .pyc and .tmp files? `bash souji` ?

source/api_documentation/transactions_api/advertiser_user.rst
```diff
@@ -25,7 +25,7 @@ The API follows REST conventions. Perform an HTTPS GET to the URL with the forma
     - Returns a JSON array of transaction objects. ``https://mynetwork.invoca.net/api/@@TRANSACTION_API_VERSION/advertisers/transactions/33.json``
 
 
-.. include:: intro.rst
+.. include:: _intro.rst
```
spencerpbartz, line 28: Thanks for cleaning this up. We only had a convention that custom templates would begin with an underscore but nothing about including other files ("partials" if you will).
Partials don't even have to be .rst files btw, but it looks good to be consistent.

spencerpbartz, line 28: Come to think of it, I think there is code in conf.py that looks specifically for files starting with underscores and identifies as something it needs to templatize. Might be worth looking into avoid unnecessary processing.

README.md
liamcardenas, line 27: `cp html souji`
`printf "\nmake souji" >> souji`

then,

`bash souji`

:)

...but 'make souji' still works too!

source/api_documentation/transactions_api/advertiser_user.rst
liamcardenas, line 28: Currently, I do not believe that there are any partials that are not .rst. Should we make it standard to make all of them .rst? My regular expression for error suppression currently assumes .rst

.gitignore
mikeweaver, line 9: Ok

README.md
spencerpbartz, line 27: Nice. Thanks for keeping that around! :-)

source/api_documentation/transactions_api/advertiser_user.rst
spencerpbartz, line 28: How about a warning if the file is _not_ .rst?

mikeweaver, line 28: We are not going to support anything other than rst.

PR #70, opened 2016-06-23
STORY-4108 Bump version to 2016-07-01

liamcardenas:
@jebentier

### DIFF
--- a/source/api_documentation/network_integration/advertiser_campaigns/index.rst
+++ b/source/api_documentation/network_integration/advertiser_campaigns/index.rst
@@ -508,3 +508,70 @@ Endpoint
    :path: /advertiser_campaigns/&lt;advertiser_campaign_id&gt;/unarchive
    :description: Unarchive a Campaign
    :page: post_advertiser_campaign_unarchive
+
+
+Error Handling
+--------------
+
+Forbidden – 403:
+
+PUT/POST
+""""""""
+
+``https://invoca.net/api/@@NETWORK_API_VERSION/<network_id>/advertiser/<advertiser_id_from_network>/advertiser_campaign/<advertiser_campaign_id_from_network>/advertiser_campaigns/<advertiser_campaign_id>.json``
+
+Content Type: application/json
+
+Response Code: 403
+
+**Request Body**
+
+.. code-block:: json
+
+  {
+    "node_type":"Menu",
+    "prompt":"Prompt text",
+    "prompt_id_from_network":"",
+    "prompt_url":null,
+    "prompt_recieved":null,
+    "children": [
+      {
+        "node_type":"Menu",
+        "prompt":"",
+        "prompt_id_from_network":"",
+        "prompt_url":null,
+        "prompt_recieved":null,
+        "children": [
+          {
+            "node_type":"EndCall",
+            "prompt":"",
+            "prompt_id_from_network":"",
+            "prompt_url":null,
+            "prompt_recieved":null
+          }
+        ]
+      }
+    ]
+  }
+
+**Response Body**
+
+.. code-block:: json
+
+  {
+    "error": {
+      "ivr_tree": {
+        "children": [
+          {
+            "0": {
+              "prompt": [
+                "cannot be empty"
+              ]
+            }
+          }
+        ]
+      }
+    }
+  }
+
+The number in error message represents the index of the child node in the tree, or in other words, it is the keypress of the node containing the error minus one.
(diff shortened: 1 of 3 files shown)


source/doc_versions.py
```diff
@@ -1,9 +1,9 @@
 # The API version strings to display in the documentation
 # COMMON_VERSION is the latest version of any one API, and is considered the overall API version
-COMMON_VERSION = '2016-06-12'
+COMMON_VERSION = '2016-07-01'
```
mikeweaver, line 3: The branch name indicates there should be more to this than just a version bump. Where is it?

liamcardenas, line 3: @mikeweaver there had previously been no documentation for this error, so there was nothing to change. In grooming we determined that if I added documentation for that one error, then we would have to document all errors, so we decided against it.

I named the branch to summarize the jira ticket (or in other words, why I'm making these changes in the first place). Should rename it?

mikeweaver, line 3: I think this is why we have internal facing versions. I don't see a reason to bump the version in the docs when there are no changes to the docs.

jebentier, line 3: we're bumping the version on the api, so don't we need to bump the version in the docs so that the docs have the new version that people can use?

mikeweaver, line 3: Not if its an internal only use case. Once we make public facing changes (that affect the docs) we can publish the new docs. In the meantime, it will simply lead to confusion in customers. "You release a new version of your API, but the docs are the same? What?!?!?"

jebentier, line 3: It's not an internal only use case though.  The changes being made to the API are public facing, just only in the structure of the error responses that come back.  We don't record the error structure on the API doc though so there are no changes made to the documentation.

nburwell, line 3: I agree this does impact the public facing API (response).

> We don't record the error structure on the API doc though so there are no changes made to the documentation.

Shouldn't we document them? Seems like a miss from before and now would be a great time to add some documentation of how errors are returned for an invalid campaign (with IVR tree)

liamcardenas, line 3: Well, this won't exactly impact or break anyone's code. If you set up your client to call invoca.com/.../**2016-05-01**/....json, then nothing will have changed for you. It's only once you call invoca.com/.../**2016-07-01**/....json, that there will be a change in behavior.

One could argue that, by not bumping the documentation, we are depriving api users of the knowledge about this new update. But since we aren't updating the documentation anyway, they wouldn't know about it either way.

Mike definitely has a point here.

mikeweaver, line 3: Yes, either we update the 7/1 docs, or we don't publish that version of the API at all.

nburwell, line 3: I'm confused why we wouldn't document the error structure for the API, that seems pretty important?  

My preference would be to update the 7/1 docs and publish the version.. [we also really need public facing release notes so that customers can quickly know what has changed in a version, but that's a separate discussion that I am trying to advocate with Nathan].

liamcardenas:
@mikeweaver @nburwell @jebentier 

I added error documentation

source/api_documentation/network_integration/advertiser_campaigns/index.rst
```diff
@@ -508,3 +508,70 @@ Endpoint
...
+                "cannot be empty"
+              ]
+            }
+          }
+        ]
+      }
+    }
+  }
+
+The number in error message represents the index of the child node in the tree, or in other words, it is the keypress of the node containing the error minus one.
```
mikeweaver, line 577: Is it always the keypress - 1? That doesn't sound right. What if I only have two key presses, 5 and 7? Won't the indexes be 0 and 1?

liamcardenas, line 577: @mikeweaver it is not possible to specify only key presses 5 & 7. Key presses must start at 1 and do not skip anything. If someone sends a request:

`{
   node_type: "Menu",
   ...
   children: [
      child1,
      child2
   ]
}`

Those children are implicitly marked as keypress 1 and 2. We are then returning the location of where to find the error in the tree that they sent, which is the same thing as keypress - 1.

mikeweaver, line 577: That is strange behavior. Seems like a common feature to have IVRs that define the first few numbers, then 9 or *. But, ok, given that, this seems fine.

nburwell, line 577: > Seems like a common feature to have IVRs that define the first few numbers, then 9 or *

That has been a feature request, yup!  I hope after this refactor is out, we can have product go back through IVR requests we've had and get a prioritized list together of high-value ones (I think this particular one would involve some updating on the model/RS side too, but now at least the UI side should be much easier)

liamcardenas:
@mikeweaver @nburwell so are we good to merge?

nburwell:
LGTM..

PR #71, opened 2016-07-13
1607/eb/story-4057 optional id from network on create

zsmathe:
@jebentier Do you mind taking a look at these API doc updates related to the id_from_network auto-generation? ([STORY-4057](https://ringrevenue.atlassian.net/browse/STORY-4057))


source/api_documentation/network_integration/advertiser_campaigns/_post_advertiser_campaign.rst
```diff
@@ -4,122 +4,24 @@
 
   .. rubric:: Examples
```
zsmathe, line 6: Removed the examples below because they were not really providing anything special over the last remaining example other than the curl command.

source/api_documentation/network_integration/advertiser_campaigns/index.rst
```diff
@@ -211,10 +223,6 @@ You are not allowed to delete campaigns.
     - integer
     - Default Creative ID.
 
-  * - max_promo_numbers
-    - integer
-    - Maximum Promo Numbers.
-
```
zsmathe, line 217: Removed this attribute because it's not to be found in the response. It was removed more than a year ago from the dto.

jebentier:
@zmathe-sv everything from a code point and an example standpoint looks really good here.  have you verified that everything looks as it should on the pages?  that's the only place we've been really screwed by.

zsmathe:
@jebentier thanks for reviewing! Yes, I verified the pages in the browser and found no glitches.

jebentier:
awesome! 👍 from me then

PR #79, opened 2016-07-25
STORY-4307 update SAPI timestamp default from Pacific to UTC

eric-dowty:
Updates SAPI timestamp documentation for compressed format to always parse in UTC.


jakesutton:
9X 👍

PR #83, opened 2016-08-25
Gk/solutionsjs updates

garvank:
@Inanepenguin please CR


source/web_integration/pnapi_web_integration.rst
```diff
@@ -68,10 +70,38 @@ Client Side Parameters
...
+      Optionally accepts a hash with the number as a key name, and campaignId as it's value like so:
+      
+      `{ "888-999-1010": 'campaignIdHere' }`
+
+  * - destinationAsId
+    - Bool
+    - Default `false`
+
+      Strips out any special characters and uses found destination phone number's digits as campaign ID.
+      Warning! Requires Invoca campaign with explicit ID for each number.
```
peterwilson: What do you think about bolding this, and possibly putting it on a separate line from the previous sentence?

source/web_integration/pnapi_web_integration.rst
```diff
@@ -95,11 +128,11 @@ Client Side Parameters
...
       These values are updated every subsequent visit.
 
-  * - savePoolParams
+  * - updatePoolParams
     - Bool
-    - Default: `false`
+    - Default: `true`
 
-      When true, the integration code will not update `poolParams` on subsequent page visits.
+      When `false`, the integration code will not update `poolParams` on subsequent page visits.
```
peterwilson, line 136: Are we actually changing parameter names and flipping the contract in the same version? Is `2016-08-01` available to customers already?

peterwilson:
@garvank should we update line 381, `<script type="text/javascript" src="//cdn.invoca.solutions/js/solutions-1.2.min.js">`, to point to latest?

peterwilson:
@garvank CR complete

source/web_integration/pnapi_web_integration.rst
nburwell: I'll let Garvan answer/clarify the above, but speaking of versions.. the PNAPI js is on a different versioning schema (2.1 ish), can you add the version number that is currently live (and therefore the documents are accurate for) on this page, probably near the top or near this parameters table?

garvank: This documentation applies to all of `2.0.0` and above, and any changes we make are backwards compatible.. That said, we did indeed flip the contract here, but actually have  compatibility for the old parameter name. Going forward, we would like customers to use this naming convention instead.

source/web_integration/pnapi_web_integration.rst
garvank: I like that idea!

garvank:
> should we update line 381 src="//cdn.invoca.solutions/js/solutions-1.2.min.js", to point to latest?

That is left at the old url intentionally. It is demonstrating how to update from an older version.

garvank:
Hey @Inanepenguin checking in on this, hoping to get it out by tomorrow if possible 😄

source/web_integration/pnapi_web_integration.rst
peterwilson: Understood, thanks!

peterwilson:
@garvank looks good :+1: :shipit:

PR #85, opened 2016-08-29
add pnapi-integration updates to 8-22

garvank:
@Inanepenguin here is the update to the latest version :/

### DIFF
--- a/source/web_integration/pnapi_web_integration.rst
+++ b/source/web_integration/pnapi_web_integration.rst
@@ -389,3 +389,4 @@ to just this:
 
   <script type="text/javascript" src="//solutions.invocacdn.com/js/pnapi_integration-latest.min.js"></script>
 
+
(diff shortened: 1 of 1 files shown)


source/web_integration/pnapi_web_integration.rst
```diff
@@ -353,4 +389,3 @@ to just this:
 
   <script type="text/javascript" src="//solutions.invocacdn.com/js/pnapi_integration-latest.min.js"></script>
 
-
```
peterwilson, line 356: I am concerned about this line removal. It wasn't in the last change, and RTD can be very very particular about new lines

peterwilson:
@garvank one concern about the new line removal. Otherwise it looks good (as did the last one!)

garvank:
Cool, thanks @Inanepenguin I agree RTD is quite fickle. Removed!

PR #86, opened 2016-09-21
TECH-1174 Update version and include caller id documentation

liamcardenas:
@jebentier


jebentier:
code looks good, just make sure to compile locally and double check that everything looks good before getting into qa.

PR #88, opened 2016-10-31
WEB-2504 add new params to the list of transaction api params

jebentier:
@mikeweaver adding the two new columns to the api docs

### DIFF
--- a/source/api_documentation/transactions_api/_advertiser_param_table.rst
+++ b/source/api_documentation/transactions_api/_advertiser_param_table.rst
@@ -173,6 +173,14 @@
     - Call Start Time (XML formatted)
     - Start of the call in Soap XML formatted time.
 
+  * - start_time_network_timezone
+    - Call Start Time Network Timezone
+    - Start of the call in the networks's time zone, followed by offset from GMT.
+
+  * - start_time_network_timezone_xml
+    - Call Start Time Network Timezone (XML formatted)
+    - Start of the call in the network's time zone in Soap XML formatted time.
+
   * - syndicated_ident
     - Syndicated ID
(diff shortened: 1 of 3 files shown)


source/api_documentation/transactions_api/_advertiser_param_table.rst
```diff
@@ -173,6 +173,14 @@
     - Call Start Time (XML formatted)
     - Start of the call in Soap XML formatted time.
 
+  * - start_time_network_timezone
+    - Call Start Time Network Timezone
+    - Start of the call in the networks's time zone, followed by offset from GMT.
```
mikeweaver, line 178: Isn't this an ISO 8601 date format? If so, we should reference that.

jebentier: I do not believe it technically is because it uses a space instead of a `T` as the separator between the date and the time portion of the timestamp.

HParker: Looks like the [spec](https://tools.ietf.org/html/rfc3339#section-5.6) allows a space.

> NOTE: ISO 8601 defines date and time separated by "T".
>   Applications using this syntax may choose, for the sake of
>   readability, to specify a full-date and full-time separated by
>   (say) a space character.

But it seems discouraged?

> By mutual agreement the parties in information interchange may transfer the date and time format
> representations. Only the date and time format representations permitted by this International Standard shall
> be used. - [2004 spec](http://dotat.at/tmp/ISO_8601-2004_E.pdf)

source: [SO:9531524](http://stackoverflow.com/questions/9531524/in-an-iso-8601-date-is-the-t-character-mandatory)

Also, I don't know why I am still a watcher on this project. Ima change that. :)

PR #95, opened 2017-01-26
1701/eb/story-4852 add signal api link and text 2016-05-01

zsmathe:


zsmathe:
@mikeweaver this is a small change that I already merged into the more recent api versions, but starting with 2016-05-01 and older I'm unable to do the merge (these are protected branches and require project collaborators).
Can you please take a look and merge this and related PRs: #96 #97 #98 #99 ?
Thank you!

PR #100, opened 2017-01-30
STORY-4754_country_advertiser_campaigns_ui_api: added campaign countr…

Vicentecarrillo:
…y to NI advertiser campaign table and examples

@mikeweaver


mikeweaver:
Should we also add country to examples for POST and PUT?

Vicentecarrillo:
I added it to POST. I don't think it makes sense to add it to the existing PUT example as it is centered around updating an IVR tree

mikeweaver:
LGTM

Vicentecarrillo:
@mikeweaver can you merge this? I don't have permissions

PR #135, opened 2018-01-10
Updated Web integration section per product & CS request

nburwell:
to point new customers over to Invoca Tags documentation in uservoice

@codyjca can you review please? I already got sign off from Clay and James


codyjca (approved):
👍

PR #136, opened 2018-01-10
Updated Web integration section per product & CS request (2017-04-01)

nburwell:
to point new customers over to Invoca Tags documentation in uservoice

@codyjca PR for 2017-04-01 version, because... read the docs.

Same commit as https://github.com/Invoca/developer-docs/pull/135


codyjca (approved):
👍

PR #137, opened 2018-02-05
Add documentation for onCompleteOverride and list argumetns for onCom…

garvank:
…plete

### DIFF
--- a/source/web_integration/pnapi_web_integration.rst
+++ b/source/web_integration/pnapi_web_integration.rst
@@ -183,7 +183,19 @@ Client Side Parameters
     - Function
     - Default: `null`
 
+      Arguments: `mapping`
+
       Name of a function to call when Invoca has finished running. Note: Do not call the function, simply pass reference.
+
+      For example: `functionName` not `functionName()`.
+
+  * - onCompleteOverride
+    - Function
+    - Default: `null`
+
+      Arguments: `batchId`, `mapping`
+
+      Name of a function to call when Invoca has finished running.
 
       For example: `functionName` not `functionName()`.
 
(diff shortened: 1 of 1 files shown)


garvank:
@nburwell Could you please review? I think all we wanted to add was `onCompleteOverride` and add the arguments it accepts. Happy to add anything else we need.

source/web_integration/pnapi_web_integration.rst
```diff
@@ -183,10 +183,22 @@ Client Side Parameters
...
 
       For example: `functionName` not `functionName()`.
 
+  * - onCompleteOverride
+    - Function
+    - Default: `null`
+
+      Arguments: `batchId`, `mapping`
+
+      Name of a function to call when Invoca has finished running.
```
nburwell: I would mention this one:

```
Name of a function to call when Invoca has finished running. If defined, the function is responsible for updating phone numbers on the web page as the default number replacement code will not be run.
```

source/web_integration/pnapi_web_integration.rst
```diff
@@ -183,10 +183,22 @@ Client Side Parameters
     - Function
     - Default: `null`
 
+      Arguments: `mapping`
+
       Name of a function to call when Invoca has finished running. Note: Do not call the function, simply pass reference.
```
nburwell: Given the `onCompleteOverride` function is now also defined, how about clarifying this one to something like:

```
Name of a function to call when Invoca has finished running.  This is called after Invoca has run the default number replacement code and swapped out any numbers that were found. Note: Do not ...
```

nburwell (commented):
Minor comments on the help copy

nburwell (approved):
Looks good

PR #142, opened 2018-02-08
1802/h/STORY-5945_2017-04-01: Resize image and fix rst file

jasonla:


nburwell (commented):
LGTM (although now in looking at this more closely.. curious why you resized image .. for it to look good on retina I think we want the original size and then to set at half width - we can sync in AM, and it's not critical for this one small image..)

nburwell (approved):
as discussed, older versions are fine with the smaller image

PR #188, opened 2018-06-21
STORY-6250: rename affiliate campaign relationship endpoint

jebentier:
After talking with Lauren, we like this name enough that we're going to backfill this name change to all previous versions.  I'll just cherrypick this commit to other versions when this is approved.

### DIFF
--- a/source/api_documentation/network_integration/advertiser_affiliate_relationships/index.rst
+++ b/source/api_documentation/network_integration/advertiser_affiliate_relationships/index.rst
@@ -1,8 +1,11 @@
-Advertiser Affiliate Relationships
+Auto-Approve Affiliate to Campaigns
 ==================================
 
 Changes to the Relationship between Advertiser and Affiliate on the network platform are replicated to the Invoca platform using this API.
 The operations on Advertiser‐Affiliate Relationships are similar to Network, in that the interface is fully idempotent, and the create and update commands will act as “create or update”.
+
+When setting the Relationship between an Advertiser and Affiliate to Approved, all current and future campaign applications made between the Affiliate and Advertiser will be auto-approved. 
+If the Relationship is set to any status other than Approved, all current and future applications will be declined and any active affiliate campaigns will be suspended.
 
 Relationship status is set individually but reading is available for one or all relationships for an advertiser.
 
--- a/source/api_documentation/network_integration/index.rst
+++ b/source/api_documentation/network_integration/index.rst
@@ -11,7 +11,7 @@ Advertiser Campaigns
 Campaign Terms
 RingPools
 Promo Numbers
-Advertiser Affiliate Relationships
+Auto-Approve Affiliate to Campaigns
 
 Terminology
 


source/api_documentation/network_integration/advertiser_affiliate_relationships/index.rst
```diff
@@ -1,4 +1,4 @@
-Advertiser Affiliate Relationships
+Auto-Approve Affilaite to Campaigns
```
nburwell: typo in `Affiliate`

source/api_documentation/network_integration/index.rst
```diff
@@ -11,7 +11,7 @@ Advertiser Campaigns
 Campaign Terms
 RingPools
 Promo Numbers
-Advertiser Affiliate Relationships
+Auto-Approve Affilaite to Campaigns
```
nburwell: typo in `Affiliate`

nburwell (changes requested):
The new name is most likely the most common use case of this end point, but are we concerned that it doesn't cover the other use cases? 

By sending 'Suspended' (or 'Deactivated') you can bulk suspend all affiliate campaigns between this affiliate and advertiser.

--- 

I think we can also improve the intro paragraph to describe a bit what this does -- ie will: approve any existing affiliate campaigns (regardless of what status they were in before), and otherwise mark this affiliate as being auto-approved anytime they apply for a campaign with this advertiser in the future.

source/api_documentation/network_integration/advertiser_affiliate_relationships/index.rst
jebentier: done

source/api_documentation/network_integration/index.rst
jebentier: done

jebentier:
Lauren and I went back and forth on that with the name and Lauren decided that the name can be this, and that the description still indicates that you can approve or deny/other actions.  I can try taking a stab at the wording for the description, but we originally talked about only the title needing to be changed

source/api_documentation/network_integration/advertiser_affiliate_relationships/index.rst
```diff
@@ -1,9 +1,12 @@
-Advertiser Affiliate Relationships
+Auto-Approve Affiliate to Campaigns
 ==================================
 
 Changes to the Relationship between Advertiser and Affiliate on the network platform are replicated to the Invoca platform using this API.
 The operations on Advertiser‐Affiliate Relationships are similar to Network, in that the interface is fully idempotent, and the create and update commands will act as “create or update”.
 
+When setting the Relationship between an Advertiser and Affiliate to Approved, all current and future campaign applications made between the Affiliate and Advertiser will be auto approved.
```
nburwell: sorry small nit, didn't notice on slack:

You have `Auto-Approve` in the title, but use `auto approved` here in the sentence. Can we be consistent on both.

jebentier: done

nburwell (approved):

caitlinrhd (approved):
Looks good. Can you update the ticket title to reflect that we didn't end up hiding this page?

jebentier:
done

PR #191, opened 2018-09-11
Remove mention of go_live via GET request from docs.

scottpchow23:


nburwell (approved):
:shipit: (once web code to support POST is deployed)

nburwell (changes requested):
Actually, in looking at the web side.. can you also remove two other GET examples:

* archive
* unarchive

They are setting a new status on a campaign and shouldn't be done via GET

nburwell (approved):
Thanks!

PR #194, opened 2018-11-08
STORY-6481 network integrations api should return all users

rjkearns:
update documentation to include new version and new field when retreiving users for an organization


source/api_documentation/network_integration/affiliates/index.rst
```diff
@@ -47,33 +47,79 @@ Similar to advertisers, you are not allowed to delete if one or more campaigns e
...
+    -
+
+  * - notify_on_weekly_campaign_performance_reports
+    - boolean, optional, defaults to false
+    -
+
+  * - notify_on_call_activities
+    - boolean, optional, defaults to false
+    -
+
```
nburwell, line 122: shouldn't you document `can_login_via_platform` ?

```
- can_login_via_platform
- boolean (read-only)
- Indicates if user can directly login with username and password. If false, the user is managed via SSO.
```

source/api_documentation/network_integration/networks/index.rst
```diff
@@ -21,33 +21,79 @@ Parameters
...
+    - boolean, optional, defaults to false
+    -
+
+  * - notify_on_weekly_campaign_performance_reports
+    - boolean, optional, defaults to false
+    -
+
+  * - notify_on_call_activities
+    - boolean, optional, defaults to false
+    -
```
nburwell, line 96: same comment here to add the new field

nburwell (changes requested):
I believe this section of pages needs to be updated to (at least the example responses):
https://github.com/Invoca/developer-docs/tree/2018-05-01/source/api_documentation/network_integration/advertiser_users

Please also make sure to add the new field to the advertiser's table too:
https://developers.invoca.net/en/2018-05-01/api_documentation/network_integration/advertisers/index.html

source/api_documentation/network_integration/affiliates/index.rst
rjkearns: I did initially but did not mark it as read-only (didn't realize that was a pattern) and removed it because it seemed like the table was for editable request params. I now see the read-only option in there as well. I'll add it back in. Thanks for catching that.

rjkearns:
@nburwell I've added the new parameter to index table but am waiting on the advertiser users endpoint section. I had a few questions that I would like sync with you/lauren on before making the changes required in web.

rjkearns:
@nburwell documentation for advertiser user endpoint has been updated

nburwell (approved):

PR #198, opened 2019-02-14
1902/story 7000 2018 02 01 sapi documentation for pce

vdinovi:
Added ` New Signal Transactions` section for SAPI 2018-02-01

### DIFF
--- a/source/api_documentation/signal_api/index.rst
+++ b/source/api_documentation/signal_api/index.rst
@@ -190,8 +190,6 @@ All examples below correspond to a date time of **11 April 2016** at **1 PM Paci
 
     **Note:** The file upload user interface includes a timezone dropdown feature, which allows users to select which timezone this format is parsed in. Use the following to format Excel date cells: **yyyy/mm/dd hh:mm:ss.000 AM/PM**.
 
-
-
 Example POST Request Using cURL
 -------------------------------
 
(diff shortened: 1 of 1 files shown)


source/api_documentation/signal_api/index.rst
```diff
@@ -470,6 +470,80 @@ Note: As signals are applied to the call, the response time of the API will incr
...
+
+Request Parameter changes:
+
+* The existing fields *partner_unique_id*, *occurred_at_time*, and *revenue* can now be specified at the top-level of a request following the same formatting described above. This will be applied to all signals and custom data in the request.
+* These fields can still be specified inline with each signal, but signals with different values for *partner_unique_id* will be not be grouped together.
+* If specified inline for a specific signal, it will take precedence for that signal over the top-level value.
+* The *revenue* field may not be specified at the top-level and inline with signals. If specified inline with signals, it must be the same for all signals with the same partner unique id. This is necessary to prevent ambiguitiy in revenue application.
+* Custom Data fields will be grouped together into a Post Call Event transaction using the top-level *partner_unique_id* (this may be omitted and will default to the empty string). 
+
+Using the same *partner_unique_id* for all signals is **highly recommended** as doing so will greatly improve the performance of the Signals API, Transactions API, and Reports.
```
nburwell: Slight re-word suggestion:

> Using a single *partner_unique_id* at the top level, such that it will be used for all signals and Custom Data, is **highly recommended.**  This will minimize the number of transactions that show up in Transaction Detail reports and the Transactions API.

source/api_documentation/signal_api/index.rst
```diff
@@ -470,6 +470,80 @@ Note: As signals are applied to the call, the response time of the API will incr
   curl -k -H "Content-Type: application/json" -X POST -d '{"search": {"transaction_id": "00000000-00000001"},"signals": [{"name": "Quality Call","value": "true"}], "custom_data": [{"name": "channel", "value": "Paid Search"}],"oauth_token": <YOUR OAUTH TOKEN>}'  https://invoca.net/api/@@SIGNAL_API_VERSION/transactions.json
   curl -k -H "Content-Type: application/json" -X POST -d '{"search": {"transaction_id": "00000000-00000001"},"signals": [{"name": "Appointment Made","partner_unique_id": "1","occurred_at_time": "1440607313","value": "false"}], "custom_data": [{"name": "channel", "value": "Paid Search"}],"oauth_token": <YOUR OAUTH TOKEN>}'  https://invoca.net/api/@@SIGNAL_API_VERSION/transactions.json
 
+New Signal Transactions
```
nburwell: A few high level comments:

I wonder if we should leave with New Signal Transactions? Or maybe we do after we have more customers live... 

Either way, there should be a corresponding section on `Updates and Idempotency` like there is above, where this mentions how it's only based on partner_unique_id. Please include a note on revenue is added when new PUID is used, versus updated on existing PUID if using the same PUID. 

Best practices above don't apply the same as down here (so maybe up on L461 indicate that is only if not on the new system?)

nburwell (commented):
nice! just a few comments

source/api_documentation/signal_api/index.rst
vdinovi: Re ordered it so the order is
... (previous docs) ...
* Signal Transactions Rollup
* Updates and Idemptotency
* Best Practices
...

Added change notices to those two sections regarding Signal Transactions Rollup

nburwell (approved):

PR #202, opened 2019-03-13
TECH-3209 signal and pce limit

gabe-kent:
- Added note regarding PCE limit
- Fixed some formatting so that the example requests show up properly

https://developers.invoca.net/en/2019-02-01/api_documentation/signal_api/index.html#signal-transactions-rollup

### DIFF
--- a/source/api_documentation/signal_api/index.rst
+++ b/source/api_documentation/signal_api/index.rst
@@ -346,6 +346,8 @@ If you are unsure about whether you are using this feature, please contact the I
 All Signals and Custom Data within a single request will be grouped together based on the value of the field **partner_unique_id**. There will be a single transaction for each group created.
 These new transactions will have the transaction type *Post Call Event*.
 
+A maximum of 10 transactions can be applied for a single request. However, this limit does not apply to the number of Signals or Custom Data included in a single grouping. See the **Best Practices** section for understanding why this limit is enforced.
+
 Request Parameter changes:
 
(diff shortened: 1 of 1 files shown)


source/api_documentation/signal_api/index.rst
```diff
@@ -346,6 +346,8 @@ If you are unsure about whether you are using this feature, please contact the I
 All Signals and Custom Data within a single request will be grouped together based on the value of the field **partner_unique_id**. There will be a single transaction for each group created.
 These new transactions will have the transaction type *Post Call Event*.
 
+A maximum of 10 transactions can be applied for a single request. However, this limit does not apply to the number of Signals or Custom Data included in a single grouping. See the **Best Practices** section for understanding why this limit is enforced.
```
vdinovi, line 349: This section is for STR users only. Is there a corresponding note for non-STR users?

vdinovi (approved):
👍, just the one question I left.

Also, just FYI -- because I had to do this same thing earlier. This change needs to be copied to all api versions >= 2018-02-01 since thats when STR was enabled. There should be a branch for each version. Just tag me on all of the PRs.

source/api_documentation/signal_api/index.rst
gabe-kent: Yes, for the most part. The Best Practices section discusses the 10 signal limit and assumes that STR is off. So I believe that covers the limit note for non-STR users. 

I assume most of the documentation is written towards non-STR users, and the STR section simply outlines the differences.

PR #226, opened 2019-10-14
2016 10 01 removed unwanted constants for affiliates

pvkadam:
Updated the Older version retaining with 2 options for Network & Advertiser Users, and updated into a single option for Affiliates


gabe-kent (approved):
Looks great!

PR #239, opened 2020-01-14
Web 4882 update email conversion reporting

annekemcgrady:
Support requested an update to the conversion reporting email documentation as it was out of date and led to customer confusion.  
You can see my changes here: https://developers.invoca.net/en/web_4882_update_email_conversion_reporting/api_documentation/conversion_reporting_email.html?next=https%3A%2F%2Fdevelopers.invoca.net%2Fen%2Fweb_4882_update_email_conversion_reporting%2Fapi_documentation%2Fconversion_reporting_email.html&ticket=ST-1579038716-NNbKFUFhs7BNc5s3ClCx7Lm9aPOj7MHW


rjkearns (commented):
This all looks fine to me, but I have two questions about the changes in this PR

1. The image you included lists an actual customer's network name, is that intentional? I would think we would want to edit that out as well.

2. This change is only being applied to the current API version, should this change be applied to older API versions?

annekemcgrady:
Hi @rjkearns - 
1) Thank you for noticing that name in the image, we missed that.  I have replaced the image with a new one with the Network name obscured.   
2) Per Luan we are only going to update the current API version because conversion reporting email is used so infrequently.

rjkearns (approved):
LGTM

PR #261, opened 2020-01-31
2019-05-01 rollup field descriptions: Append rollup messages to signal tx rollup affected fields

omeedrj:
Given that these descriptions are being added in order to help other aspects of the platform as it currently is (such as the webhooks and the UI), I'd argue it's not worth updating past versions of the API docs.

Also please point out any other fields that would benefit from appended rollup related descriptions.

### DIFF
--- a/source/api_documentation/transactions_api/_affiliate_param_table.rst
+++ b/source/api_documentation/transactions_api/_affiliate_param_table.rst
@@ -43,7 +43,7 @@
 
   * - call_source_description
     - Source
-    - Source of the transaction |rollup_affects_call_message|
+    - Source of the transaction
 
   * - city
     - City
--- a/source/api_documentation/transactions_api/_affiliate_param_table.rst
+++ b/source/api_documentation/transactions_api/_affiliate_param_table.rst
@@ -1,4 +1,4 @@
-
+.. include:: _signal_transaction_rollup_replacements.rst
 
 ..  list-table::
   :widths: 30 8 40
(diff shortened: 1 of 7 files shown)


source/api_documentation/transactions_api/_affiliate_param_table.rst
```diff
@@ -35,15 +35,15 @@
...
 
   * - call_result_description_detail
     - Call Result
-    - Status of the transaction
+    - Status of the transaction |rollup_affects_call_message|
 
   * - call_source_description
     - Source
-    - Source of the transaction
+    - Source of the transaction |rollup_affects_call_message|
```
vdinovi: Was call_source affected? This just indicated where the call came from right?

source/api_documentation/transactions_api/_signal_param_table.rst
```diff
@@ -16,11 +16,11 @@ Most of the fields in this table are now deprecated. See Custom Data & Signal Pa
 
   * - signal_name *(deprecated)*
```
vdinovi, line 17: Isn't this being renamed to signal_names?

vdinovi (approved):
looks good, just two questions

rjkearns (approved):

source/api_documentation/transactions_api/_affiliate_param_table.rst
omeedrj: Good point. Thank you for pointing this out. I'll remove this.

source/api_documentation/transactions_api/_signal_param_table.rst
omeedrj: `signal_names` is available for webhooks, but not for anything else yet. I'm going to leave this as is until all other use cases have also been renamed.

vdinovi (approved):

PR #270, opened 2021-08-06
STORY-9864 updated transactions_api docs to read marketing data

laurenhedberg:
https://invoca.atlassian.net/browse/STORY-9883

* Updated `Custom Data` docs to read `Marketing Data`
* Renamed file to `_marketing_data_params.rst`
* Updated references to `Custom Data` docs to `Marketing Data`

Note: I had some difficulties building the html for these docs locally.  Looks like we are using Python 2.7 (which is being deprecated) to create the virtual environment so this caused some trouble.
`sudo easy_install pip` did not work for me because it was attempting to install a later version that did not not support Python 2.7.

I had to run the following commands instead:
`rm -f /usr/local/bin/pip`
`sudo easy_install pip==20.3.4`
After this I was able to create a virtual environment and build the docs.

In the process of debugging I also noticed we are using an old version of Sphinx. Commands in our `config.py` setup have been deprecated for years. I'm curious if we will be upgrading our Python and/or Sphinx versioning for our documentation any time soon?

* I tried building the latest version on Read the Docs but it didn't pick up my changes. I believe this needs to be merged first

## Checklist

### DIFF
--- a/source/api_documentation/transactions_api/_custom_data_params.rst
+++ b/source/api_documentation/transactions_api/_custom_data_params.rst
@@ -1,17 +1,17 @@
 
 
-Custom Data & Signal Parameters
+Marketing Data & Signal Parameters
 **********************
 
-Please click  `this link <https://www2.invoca.net/customer_data_dictionary/home>`_ to access the custom data parameters specific to your account.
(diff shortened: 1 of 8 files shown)


nburwell (approved):
Thanks for the update!

PR #273, opened 2021-09-02
STORY-10029: Add transcript API documentation page

zmjohnson42:
[JIRA TICKET](https://invoca.atlassian.net/browse/STORY-10029)

This documentation update is focusing on our new Transcript API that details how to access the transcripts for a given call. As part of this change we are adding a new primary navigation section "Call Api" and within that adding the "Transcripts" page. We added this new section as looking forward we will also want to add the "Recording Api" section in this as well, but that is more future work.

The Transcript API itself is fairly simple and modeled the page similar to the Transcation API but using the expandable API rows to display examples of our different Query Parameters. Currently our API only has one required query parameter `transcript_format` and only returns in JSON format, so a lot of the tables have been reduced to match the expectations of the API.

To confirm that this documentation is correct the corresponding tickets that add in the functionality for this update can be found here.
- https://invoca.atlassian.net/browse/STORY-10028

## Checklist

### DIFF
--- a/source/api_documentation/call_api/_get_call_transcript_agent.rst
+++ b/source/api_documentation/call_api/_get_call_transcript_agent.rst
@@ -1,0 +1,23 @@
+.. container:: endpoint-long-description
+
+  .. rubric:: Examples
+
+  Get transcript for the given call record `CUID-33` with just the agent included
+
+  Endpoint:
+
+  ``https://mynetwork.invoca.net/call/transcript/CUID-33?transcript_format=agent``
+
+  Format: application/json
+
+  Response Code: 200
+
+  Response Body:
+
+  .. code-block:: json
+
+     [
+       {
+        agent: "Which Color?, Okay I will look into that, Goodbye!"
+       }
+     ]
--- a/source/api_documentation/call_api/_get_call_transcript_agent.rst
+++ b/source/api_documentation/call_api/_get_call_transcript_agent.rst
@@ -16,8 +16,8 @@
 
   .. code-block:: json
 
-     [
-       {
+    [
+      {
         agent: "Which Color?, Okay I will look into that, Goodbye!"
-       }
-     ]
+      }
+    ]
--- a/source/api_documentation/call_api/_get_call_transcript_caller.rst
+++ b/source/api_documentation/call_api/_get_call_transcript_caller.rst
@@ -15,9 +15,9 @@
   Response Body:
 
   .. code-block:: json
-
-     [
-       {
+  
+    [
+      {
         caller: "Orange, Awesome thank you, Have a good one"
-       }
-     ]
+      }
+    ]
--- a/source/api_documentation/call_api/_get_call_transcript_caller.rst
+++ b/source/api_documentation/call_api/_get_call_transcript_caller.rst
@@ -1,0 +1,23 @@
+.. container:: endpoint-long-description
+
+  .. rubric:: Examples
+
+  Get transcript for the given call record `CUID-33` with just the caller included
+
+  Endpoint:
+
+  ``https://mynetwork.invoca.net/call/transcript/CUID-33?transcript_format=caller``
+
+  Format: application/json
+
+  Response Code: 200
+
+  Response Body:
+
+  .. code-block:: json
+
+     [
+       {
+        caller: "Orange, Awesome thank you, Have a good one"
+       }
+     ]
--- a/source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
+++ b/source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
@@ -17,8 +17,6 @@
   .. code-block:: json
 
      [
-       {
-        agent: "Which Color?, Okay I will look into that, Goodbye!",
-        caller: "Orange, Awesome thank you, Have a good one"
-       }
+       { agent: "Which Color?, Okay I will look into that, Goodbye!" },
+       { caller: "Orange, Awesome thank you, Have a good one" }
      ]
--- a/source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
+++ b/source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
@@ -15,8 +15,8 @@
   Response Body:
 
   .. code-block:: json
-
-     [
-       { agent: "Which Color?, Okay I will look into that, Goodbye!" },
-       { caller: "Orange, Awesome thank you, Have a good one" }
-     ]
+  
+    [
+      { agent: "Which Color?, Okay I will look into that, Goodbye!" },
+      { caller: "Orange, Awesome thank you, Have a good one" }
+    ]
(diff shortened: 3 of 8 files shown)


source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
```diff
@@ -0,0 +1,24 @@
...
+  Format: application/json
+
+  Response Code: 200
+
+  Response Body:
+
+  .. code-block:: json
+
+     [
+       {
```
laurenhedberg: So in cases like this, we will actually get the caller and agent in different objects. Sorry I didn't include examples yet. So it would look like this instead: 

```
[
   { agent: "Which Color?, Okay I will look into that, Goodbye!" }
   { caller: "Orange, Awesome thank you, Have a good one"  }
]
```

source/api_documentation/call_api/_get_call_transcript_caller_agent_conversation.rst
```diff
@@ -0,0 +1,28 @@
...
+  Format: application/json
+
+  Response Code: 200
+
+  Response Body:
+
+  .. code-block:: json
+
+     [
+       {
```
laurenhedberg: Same for this example, we pass back each line of text as it's own hash object. See example above for reference.

source/api_documentation/call_api/transcript_api.rst
```diff
@@ -0,0 +1,74 @@
+#####################
+Transcripts
+#####################
+
+.. include:: _rollup_message.rst
+
+URL
+---
+
+The API follows REST conventions. Perform an HTTPS GET to the URL with the format in which you’d like to receive data. The following response formats are supported, where CUID-33 is the advertiser id.
```
laurenhedberg: I think you mean `call record id` instead of `advertiser id`

source/api_documentation/call_api/index.rst
```diff
@@ -0,0 +1,20 @@
+Call API
+================
+
+The Call API is accessible using the API credentials generated on the platform.
+
+.. raw:: html
+
+The data returned depends on your account type. To see which fields you will see, select an account type below.
```
laurenhedberg: I'm not sure about the relevance of account type in the description here. We are looking at either recordings or transcripts, right? So request type may be better here

laurenhedberg (changes requested):
Thanks for putting this together! I left some suggestions on cleaning up some wording.  Some of the examples are a little off (my bad for not making those clear). I gave an example of what the data we will return should look like.

Do you mind running this one by Nash as well? It would help to take screenshots of the rendered pages. He might have some more suggestions on the wording.

source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
nburwell: Just checking.. is that desired, @laurenhedberg ? The DT API format matches what is shown currently in this api doc I believe (a single block with `agent` and `caller` as keys).

source/api_documentation/call_api/_get_call_transcript_caller_agent_conversation.rst
nburwell: Depending on answer to above.. I wonder if we instead want an array of single hashes?

```
[
  { 
    agent: "Hi",
    caller: "Yo"
  },
  {
    agent: "How can I help you?",
    caller: "Actually, wrong number"
  },
  {
    agent: "Okay. Bye"
  }
]
```

source/api_documentation/call_api/index.rst
```diff
@@ -0,0 +1,20 @@
...
+
+:doc:`transcript_api`
+
+Please note, the documentation shows a superset of fields available. Your response may differ depending on the
+features enabled on the platform being accessed. Make a sample request to the API to identify the fields returned.
```
nburwell: I think this part may be copied from the transcription api doc page as well? Do we need these sentences?

nburwell (commented):
Just a couple questions from me

source/api_documentation/call_api/index.rst
zmjohnson42: Ah correct, was grabbing from that doc when starting out, I will fix this up!

source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
laurenhedberg: This is the behavior that was confirmed with Nash, but I will double check with him on this one.

source/api_documentation/call_api/_get_call_transcript_caller_agent_conversation.rst
laurenhedberg: Confirmed with Nash, we do want to keep these caller and agent blocks separate. We know it's not perfectly aligned with existing DT format but he prefers the lines in individual blocks since it seems more consistent with a conversation.

source/api_documentation/call_api/_get_call_transcript_caller_agent_block.rst
laurenhedberg: Confirmed we will keep the existing format providing caller and agent in their own blocks

source/api_documentation/call_api/_get_call_transcript_caller_agent_conversation.rst
laurenhedberg:
```[
  { 
    agent: "Hi"
  },
  {
    caller: "Yo"
  },
  {
    agent: "How can I help you?"
  },
  {
    caller: "Actually, wrong number"
  },
  {
    agent: "Okay. Bye"
  }
]

**Review by nburwell — APPROVED** [review]

File: source/api_documentation/call_api/_get_call_transcript_agent.rst [hunk 6]
```diff
@@ -0,0 +1,23 @@
...
+
+  Response Code: 200
+
+  Response Body:
+
+  .. code-block:: json
+
+     [
+       {
+        agent: "Which Color?, Okay I will look into that, Goodbye!"
```
nburwell, line 21: super minor nit: but looks like this is only indented 1 space from `{`, and should be 2 spaces.

source/api_documentation/call_api/_get_call_transcript_caller.rst
```diff
@@ -0,0 +1,23 @@
...
+
+  Response Code: 200
+
+  Response Body:
+
+  .. code-block:: json
+
+     [
+       {
+        caller: "Orange, Awesome thank you, Have a good one"
```
nburwell, line 21: same here: should be 2 spaces for indent

lmaragon (approved):
✅

source/api_documentation/call_api/_get_call_transcript_caller.rst
zmjohnson42: Ah good catch! The spacing was off in all the examples so I have updated them to be 2 spaces for an indent!

laurenhedberg (approved):
Looks great, thank you for doing this!

One more thing actually, we will want to update the version in `doc_versions.py` to reflect the latest changes.

PR #284, opened 2022-03-01
KB-405: marking conversion reporting pages as deprecated

nburwell:
* Also removed Conversion Reporting Pixel and Conversion Reporting Email from left index (they are linked from the Conversion Reporting API page now)
* Indicated RingPool API as legacy and moved below Bulk RingPool API (the preferred way to get dynamic allocation)

Plan was to only do these changes on the latest version of docs, but open to also backporting this to previous versions if we think that is best.

Demo video:
https://drive.google.com/file/d/1DyH83o6hh5MsNyP_N2lGsJsjZzbDcowF/view?usp=sharing

Preview URL:
https://developers.invoca.net/en/2020-10-01_deprecate_conv_reporting/api_documentation/index.html

## Checklist

- [x] Test the documentation changes on readthedocs as a private branch

### DIFF
--- a/source/api_documentation/conversion_reporting.rst
+++ b/source/api_documentation/conversion_reporting.rst
@@ -1,11 +1,29 @@
-Conversion Reporting
+Conversion Reporting (Deprecated)
 ====================
+
+.. raw:: html
+
+  <div style="margin-bottom: 20px; padding: 15px; background-color: #fff3cd; color: #333;"><p>
+    <b>Notice:</b>
+    <a href="https://developers.invoca.net/en/latest/api_documentation/signal_api/index.html">Signal API</a> is now the supported way for reporting that a sale or other post call event occurred on calls.
+    The Conversion Reporting API is deprecated and no longer available for new accounts.</p>
+  </div>
 
 The Conversion Reporting API (formerly referred to as the “Call Center” or “Sales Reporting” API) is used to report completed order information (or other conversion events) from a server back into the platform.
 Reporting order information through the API allows advertisers to compensate publishers for sales that convert over the phone (or on the web).
 Additionally, Advertiser Direct campaigns using the Conversion Reporting API with search campaigns can report back revenue generated for each phone call so that search bids can be optimized.
 
-Before using the Conversion Reporting API, request credentials from questions@invoca.com
+For existing customers looking for deprecated documentation on the conversion pixel and email approach to reporting conversions, see these pages:
+
+:doc:`conversion_reporting_email`
+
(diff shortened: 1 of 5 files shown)


source/api_documentation/conversion_reporting.rst
```diff
@@ -1,11 +1,29 @@
-Conversion Reporting
+Conversion Reporting (Deprecated)
 ====================
 
+.. raw:: html
+
+  <div style="margin-bottom: 20px; padding: 15px; background-color: #fff3cd; color: #333;"><p>
+    <b>Notice:</b>
+    <a href="https://developers.invoca.net/en/latest/api_documentation/signal_api/index.html">Signal API</a> is now the supported way for reporting conversions on calls.
```
vdinovi: > now the supported way for reporting conversions

Conversions themselves are being deprecated in favor of signals right? Would it be clearer to omit _conversions_ from this to prevent confusion? Or would that be more confusing?

vdinovi (approved):
Nice!

source/api_documentation/conversion_reporting.rst
nburwell: The transaction type of "Reported Conversion" is going away, yes. I was using this word from the customer use case perspective where you can have a signal named "Post Call Conversion" or "Sale Occurred" or "Appointment Occurred" (versus 'appointment booked' that would happen on the phone call) to track "conversions". 

But, given this is the technical doc portal, if the term is too overloaded, we could go with something else like "reporting that a sale or other post call event occurred on calls." ?

vdinovi: My concern was just someone reading that thinking we are continuing to support "Conversion". I think its fine as is if you don't think that's an issue. Otherwise, yeah Sale/PCE seems good too.

nburwell: Less chance for confusion sounded good to me, so I updated the message: fd67511

kamerynn (approved):
👏

ryanbales:
nice work @nburwell ty for helping drive deprecation forward. LGTM

ryanbales (approved):

nburwell:
I ended up squashing down to one commit (which, I should have done before CR -- there were no conflicts, it was all automated, so I am confident the contents of the diff are still the same). Reason for 1 clean commit is in case we want to cherry-pick this to previous doc versions. 

And @ryanbales just FYI, once I merge this in (happening now), please be sure that you get these latest changes into your WIP new version branch. Thanks!

PR #304, opened 2022-10-20
Story 12292 sms messaging api documentation.fz1

gfaza:
[STORY-12550](https://invoca.atlassian.net/browse/STORY-12550)

This is automation for getting from rspec-openapi yaml to swagger-ui json.  works in concert with https://github.com/Invoca/sms-messaging/pull/150 to automate documentation.

## Checklist


requirements.pip
```diff
@@ -2,3 +2,5 @@ sphinx
 sphinx-rtd-theme
 sphinxcontrib-httpdomain
 -e git+https://github.com/rtfd/recommonmark.git#egg=recommonmark
+PyYAML
+ipdb
```
bradmurray-dt: NIT: no newline at end of file, isn't requirements.txt the standard naming convention for these in python?

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
```diff
@@ -0,0 +1,172 @@
...
+        parameter_references = json_dict.get('components', {}).get('parameters', {}).keys()
+
+        for path in json_dict['paths'].keys():
+            print "path: {}".format(path)
+            current_path = json_dict['paths'].pop(path)
+
+            for verb in current_path.keys():
+                print "verb: {}".format(verb)
+                try:
+                    summary = current_path[verb].get('summary')
```
bradmurray-dt: I don't think get does what you want here.  It is intended to allow you to pull a key from a dict and respond with a default or None without throwing a KeyError.  If I'm reading this right, you would never be able to hit the exception.

current_path[verb]['summary'] would throw a keyerror when missing.

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
```diff
@@ -0,0 +1,172 @@
...
+                except KeyError:
+                    warnings_missing_summary.append("{} {}".format(verb, path))
```
bradmurray-dt: NIT: the formats should probably be replaced with f-strings or simple string concatenation...
`"{} {}".format(verb, path)`
Could be:
`f"{verb} {path}"`
or 
`verb + ' ' + path`

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
```diff
@@ -0,0 +1,172 @@
...
+                    continue
+
+                summary_path = summary.split(' ')[1]
+                if summary_path not in json_dict['paths'].keys():
```
bradmurray-dt: No need to use "keys".. e.g.
`if summary_path not in json_dict['paths']`

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
```diff
@@ -0,0 +1,172 @@
...
+        if parameter_reference not in [pr.items()[0][1] for pr in current_path[verb].setdefault('parameters', [])]:
+            current_path[verb]['parameters'].append({"$ref": parameter_reference})
+
+
+for yaml_path in yaml_paths:
+    with open(yaml_path, 'r') as yaml_file:
+        json_dict = yaml.load(yaml_file.read().rstrip(), Loader=yaml.FullLoader)
+        parameter_references = json_dict.get('components', {}).get('parameters', {}).keys()
+
+        for path in json_dict['paths'].keys():
```
bradmurray-dt: No need for keys... looping through a dict will by default give you keys at each iteration.

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
```diff
@@ -0,0 +1,172 @@
...
+        if parameter not in parameter_references + warnings_undefined_path_parameter:
+            warnings_undefined_path_parameter.append(parameter)
+
+        # add the reference to the shared parameter definition
+        parameter_reference = "#/components/parameters/{}".format(parameter)
+        if parameter_reference not in [pr.items()[0][1] for pr in current_path[verb].setdefault('parameters', [])]:
+            current_path[verb]['parameters'].append({"$ref": parameter_reference})
+
+
+for yaml_path in yaml_paths:
```
bradmurray-dt: Be careful when doing work in the global scope.  It can lead to hard-to-track bugs, as variables defined here can be affected by things inside function calls and vice versa.

source/_static/css/custom.css
```diff
@@ -610,4 +610,10 @@ div.endpoint-list div.panel-body .container {
   padding: 0;
 }
 
+.response-col_description__inner,
+.response-control-media-type,
+.response-col_links
+{
+  display: none;
+}
```
cgaroutte: I have been putting any css rules that target the swagger-ui specifically into `css/swagger-ui/index.css` and targeting them with `.swagger-ui` ie 

```suggestion
.swagger-ui .response-col_description__inner,
.swagger-ui .response-control-media-type,
.swagger-ui .response-col_links
{
  display: none;
}
```

That way we don't accidentally target any other html elements outside of the embedded swagger ui

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
gfaza: this change worked for the `if` statements 👍

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
gfaza: most of the `for` didn't like removing the `.keys()`, due to `RuntimeError: dictionary changed size during iteration`.  something about the `keys()` getting us a live view at the dict gets us past this error.

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
gfaza: agreed.  between this comment and the gentle reminder of the existence of pylint, went ahead and tightened the code away from global scopes. 👍

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
gfaza: excellent catch!  i'd outsmarted myself by trying to do both 😞

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
bradmurray-dt: > most of the `for` didn't like removing the `.keys()`, due to `RuntimeError: dictionary changed size during iteration`. something about the `keys()` getting us a live view at the dict gets us past this error.

Yes, that would be an issue if you try to modify the set of keys in the dict during the execution of a  loop.

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
gfaza: locally, i was being told that f-strings are not available at Python 2.7X that this repo requires.  i was however table to update them to kwargs, which is more legible and a step closer to f-strings, once/if/when we try and port this to Python 3.X.

gfaza:
sorry/not-sorry, brad - your feedback inspired quite a bit of structural changes, though hopefully for the better!  ready for re-review, if you please.  thank you!!!

scripts/rspec-openapi-yaml-to-swagger-ui-json.py
bradmurray-dt: Yes, f-strings were introduced in 3.6  Supporting both 2.x and 3.x is tricky.  If that is really necessary, you will want to both pylint and test your script on both.

cgaroutte (changes requested):
Thanks for doing this. I can't speak much to the python code at this point, but looks like Brad is doing that. 😃 

Food for thought: was the script implemented in Python because of developer docs? Would it make sense to implement this in ruby in sms-messaging in the future since we are eventually going to move off of developer docs?

gfaza:
@cgaroutte - i can followup on that surprising `_json:` element that we see on some request body payloads as a bugfix for this.  would you be ok if I merged this into your branch, and tackled it there?  want to get the birds looking over this stuff, but would be best to do it on a single branch.

gfaza:
@bradmurray-dt, you got any more advice before I press the big green button?  wanna get this merged to Chris's branch, so the birds can review the consolidated changes before we merge to mainline.

actually - I see it's past business for you.  I'll handle any further feedback on Chris's branch, as a follow on.

PR #306, opened 2022-11-04
STORY-12292 avoiding odd _json key for request body arrays

gfaza:
[STORY-12292](https://invoca.atlassian.net/browse/STORY-12292)
Working around swaggers affinity for `_json` elements for request body json arrays


gfaza:
cc: @cgaroutte

scripts/rspec_openapi_yaml_to_swagger_ui_json.py
```diff
@@ -111,6 +111,28 @@ def process_paths():
...
+        process_request_body_arrays()
+
+    def process_request_body_arrays():
+        """Adjusts a surprising way that openapi/swagger seem to describe request body json arrays"""
+        for path in yaml_dict['paths']:
+            try:
+                request_body_json = yaml_dict['paths'][path]['post']['requestBody']['content']['application/json']
+                request_body_json['schema']['type'] = 'array'
+                request_body_json['schema'].pop('properties')
+                if re.match(r"^\/delivery_reports$", path):
```
bradmurray-dt: I may be reading this wrong, but if there are no character ranges, wildcards, etc and the regex defines both starts with (^) and ends with ($)... wouldn't a simple comparison (==) do the same thing here?

gfaza: great call!  i went into it thinking i was going to have to match on a pattern, before realizing the `path` at this point in the code was devoid of the date/version string i was anticipating from the route.

cgaroutte (changes requested):
POST body format looks perfect now. 
I agree we should hide the schemas section. That can be done by adding the following option to the swagger config in swagger-initializer.js
`defaultModelsExpandDepth: -1`

bradmurray-dt (approved):
LGTM

cgaroutte (approved):
LGTM

PR #340, opened 2023-10-19
STORY-16374 Create Endpoint for retrieving all Destinations & Campaigns

abhig10:
[STORY-16374 Create Endpoint for retrieving all Destinations & Campaigns](https://invoca.atlassian.net/browse/STORY-16374)

## Checklist


SomeSayKosm (commented):
I think the content looks good here, but I think the formatting, on what I think are meant to be tables is not quite right.  This is what the index page looks like for me.  
image  I also think it's probably worth adding something about the `include_inactive` for `_get_customer_phone_numbers.rst` page.

byendan (commented):
I'm not too familiar with this area, or python for that matter, but I do agree with Carson, the content is looking good, but the formatting could use some tweaking.

Again, I'm going to leave as just a comment so that I don't block this review while out of office.

source/api_documentation/network_integration/customer_phone_numbers/index.rst
```diff
@@ -0,0 +1,110 @@
...
+    - string
+    - Phone number of the Destination.
+
+  * - primary_url
+    - string
+    - ToDo
+
+  * - requested_destination_name
+    - integer
+    - ToDo
```
abhig10: https://github.com/Invoca/developer-docs/pull/340#pullrequestreview-1688509027 Cont.

Not sure what `primary_url` or `requested_destination_name` mean. I couldn't find a UI input related to it either. So we can either remove these from here and API as well or I can put any description that you have in mind.

abhig10:
@SomeSayKosm Thank you for catching that. This was something I wanted to discuss with you but then lost track of it. Started this above as a [comment conversation](https://github.com/Invoca/developer-docs/pull/340#discussion_r1368498525) so we can talk about it.

> I think the content looks good here, but I think the formatting, on what I think are meant to be tables is not quite right. This is what the index page looks like for me.

abhig10:
Team cc @Invoca/campaigns-and-attribution

source/api_documentation/network_integration/customer_phone_numbers/index.rst
SomeSayKosm: Yeah, there is some awkwardness in trying to find how these values map to the frontend.  

`primary_url` is the also referred to as "URL Found On" or something like that.  It marks the exact URL that the number can be found on the customer's website.  We don't really process it for anything as far as I'm aware, but the Destination Settings can be configured to allow our tag to capture CustomerPhoneNumbers automatically, in which case, we capture the `primary_url` for them, so they know where to find it if they're interested in knowing how it was generated.  

`request_destination_name` is just the "Name" or "Friendly Name" of a CustomerPhoneNumber, just an alias for identifying them beyond the phone number itself.

Screenshot 2023-10-23 at 11 38 14 AM

In my above screenshot, the URL column is empty since I manually entered them through the UI but that _would_ show the URL that the number was one if their tag was setup to detect & create destinations automatically.  In this context, I think it correct to return both values, even if they might not be that useful for FindLaw, just because it's appropriate for broad usage.

SomeSayKosm (approved):
The primary table looks better now but there's still some funkiness with the docs link at the top and what looks like tables, at the bottom.  I'm going to pre-approve since these are pretty minor, but they should probably be cleaned before being pushed out.
> I'm not sure why it looks that way for you but it looks correctly when I build the [html following the instructions mentioned here](https://github.com/Invoca/developer-docs/#getting-started).

Yeah, I decided to look back for other options and I realized that whatever variation of markdown we're using here, simply doesn't work in Github, but is correct for our docs page, so I take back what I said.  Look good!

abhig10:
I'm not sure why it looks that way for you but it looks correctly when I build the [html following the instructions mentioned here](https://github.com/Invoca/developer-docs/#getting-started).

Screenshot 2023-10-25 at 7 41 51 PM
Screenshot 2023-10-25 at 7 42 02 PM

abhig10:
@omeedrj Would you have the required permissions to merge and deploy this?

PR #348, opened 2023-11-07
STORY-16627 Add documentation for JS Tags API to developer docs

abhig10:
[Add documentation for JS Tags API to developer docs](https://invoca.atlassian.net/browse/STORY-16627)

Screenshot 2023-11-07 at 6 20 21 PM
Screenshot 2023-11-07 at 6 20 33 PM
Screenshot 2023-11-07 at 6 20 42 PM
Screenshot 2023-11-07 at 6 28 54 PM
Screenshot 2023-11-07 at 6 29 04 PM
Screenshot 2023-11-07 at 6 21 14 PM

Specific endpoint documentation

Screenshot 2023-11-07 at 6 26 34 PM
Screenshot 2023-11-07 at 6 26 44 PM
Screenshot 2023-11-07 at 6 26 54 PM
Screenshot 2023-11-07 at 6 27 05 PM
Screenshot 2023-11-07 at 6 31 20 PM
Screenshot 2023-11-07 at 6 31 28 PM
Screenshot 2023-11-07 at 6 31 35 PM
Screenshot 2023-11-07 at 6 31 42 PM

## Checklist


abhig10:
Team tag @Invoca/campaigns-and-attribution

source/api_documentation/network_integration/js_tags/_get_js_tag.rst
```diff
@@ -0,0 +1,156 @@
...
+  Endpoint:
+
+  ``https://invoca.net/api/@@NETWORK_API_VERSION/<network_id>/js_tags/<js_tag_id>.json``
+
+  Response Code: 200
+
+  Response Body:
+
+  .. warning::
+      The provided documentation and JSON response for JS Tag Revision documentation might change in the next release.
```
abhig10, line 18: Added these warnings throughout as the JS Tag Revision tickets https://invoca.atlassian.net/browse/STORY-16628 and https://invoca.atlassian.net/browse/STORY-16629 will add new API version and that wouldn't have as many details.

source/api_documentation/network_integration/js_tags/index.rst
```diff
@@ -0,0 +1,340 @@
...
+
+Revision for the JS Tag
+
+.. warning::
+    The provided documentation and JSON response for JS Tag Revision documentation might change in the next release.
+
+.. list-table::
+  :widths: 15 15 50
+  :header-rows: 1
+  :class: parameters
```
abhig10: Added the parameters here because we are returning revisions in all APIs. It might be more feasible to move these parameters to revisions documentation and add a link here.

source/api_documentation/network_integration/js_tags/index.rst
```diff
@@ -0,0 +1,340 @@
...
+    - string
+    - Friendly name for the marketing data field.
+
+  * - default_campaign
+    - object
+    - Default campaign for the all numbers found except for ones in `numbers_to_replace` and `numbers_to_not_replace` configurations.
+
+  * - default_campaign.default_campaign_id
+    - integer
+    - Internal campaign id of the default campaign. TBD: SHOULD THIS BE UPDATED TO id_from_network?
```
abhig10: This returns internal id of campaigns here, but campaign API uses `advertiser_campaign_id_from_network`. I think we should update this for 2022 API version while keeping 2017 version as is. Thoughts?

source/api_documentation/network_integration/js_tags/index.rst
```diff
@@ -0,0 +1,340 @@
...
+
+  * - default_campaign.default_campaign_name
+    - string
+    - Name of the default campaign
+
+  * - destination_param
+    - string
+    - WHAT IS THIS?
```
abhig10: I don't have a good understanding of this parameter. Can anyone provide a 1 or 2 line description to be here added here?

source/api_documentation/network_integration/js_tags/index.rst
SomeSayKosm: Yeah, I agree with your assessment.  Flipping between internal & external campaign ID would only serve to confuse.

source/api_documentation/network_integration/js_tags/index.rst
SomeSayKosm: When our JsTag sends a request to PNAPI for destination, it will store the destination phone number (that is the phone number that exists on our Customer's web page, and will correspond to a CustomerPhoneNumber) beneath this param in the request.  

Admittedly, I would need to do some more research to understand why this param is changeable and not just a static thing that we set.  

For reference, [here](https://github.com/Invoca/web/blob/0434e693536df63ea6c63549652904d27fefa8b9/lib/js_tag/code_generator.rb#L31) is where it's referenced in web to be packaged into the JsTag revision that will be saved in S3. 
And [here](https://github.com/Invoca/PNAPI/blob/bc07a373d1210f38dd2e36a137fe413728be2bca/lib/pnapi/managers/pool_manager.rb#L45) it is being pulled from the request in PNAPI to be used for a RingPool look.

source/api_documentation/network_integration/js_tags/index.rst
SomeSayKosm: I think it's fine to leave the documentation here as well.  Keeps users from needing to tab to another endpoint's page despite only using this one.

SomeSayKosm (commented):
I think this looks stellar.  Just holding approval for your idea of updating the version and incorporating the id_from_network

source/api_documentation/network_integration/js_tags/index.rst
abhig10: I wanted to avoid duplication. Let me see if I can create include it using a file.

source/api_documentation/network_integration/js_tags/index.rst
abhig10: 👍 Will be done in https://invoca.atlassian.net/browse/STORY-16628 and https://invoca.atlassian.net/browse/STORY-16629 since those tickets will actually work on revisions.

source/api_documentation/network_integration/js_tags/index.rst
abhig10: It is possible to include files. Moved it to a new file that's now included in JS tags documentation.

source/api_documentation/network_integration/js_tags/index.rst
abhig10: Added: "Destination key for dynamic routing."

source/api_documentation/network_integration/js_tags/index.rst
```diff
@@ -0,0 +1,182 @@
...
+
+Manage Invoca Tags (aka JS Tags)
+""""""""""""""""""""""""""""""""""""""""""""""""
+
+The Invoca Tag is a snippet of code that connects Invoca to your landing pages. This API documentation helps provide necessary features to create and manage an Invoca Tag and it's revisions.
+
+.. raw:: html
+
+    <p>
+    More details on the Invoca Tags can be found <a href="https://community.invoca.com/t5/call-tracking/how-to-configure-and-deploy-your-invoca-tag-with-the-invoca-tag/ta-p/465#U465">here</a> and <a href="https://community.invoca.com/t5/developer-features/invoca-tag-wizard-custom-code-settings/ta-p/691">here</a>.
```
byendan: For the links I think it would be good to format it a little differently, and show the title of the articles. I'm imagining a bulleted list with the links

byendan (changes requested):
Giving some early feedback to make sure I'm not blocking you longer than needed. 
I still need to take a closer look at the specific endpoint docs.

source/api_documentation/network_integration/js_tags/index.rst
abhig10: Done in bc7334279036ff8e8f04d57911e997d29e08bfc8

Screenshot 2023-11-09 at 12 11 37 AM

source/api_documentation/network_integration/js_tags/js_revision_parameters.rst
```diff
@@ -0,0 +1,168 @@
...
+
+  * - default_campaign
+    - object
+    - Default campaign for the all numbers found except for ones in `numbers_to_replace` and `numbers_to_not_replace` configurations.
+
+  * - default_campaign.default_campaign_id
+    - integer
+    - Internal campaign id of the default campaign.
+
+      TBD: SHOULD THIS BE UPDATED TO advertiser_campaign_id_from_network.
```
byendan, line 82: I believe these map to the id_from_network, updating a campaign's id_from_network requires an update to the tag to make sure it continues to map to that campaign.

source/api_documentation/network_integration/js_tags/js_revision_parameters.rst
```diff
@@ -0,0 +1,168 @@
...
+
+  * - numbers_to_replace[].phone_number_e164
+    - string
+    - E.164 version of the phone number to be replaced.
+
+  * - numbers_to_replace[].advertiser_campaign_id
+    - integer
+    - Id of the advertiser campaign to be used for the above phone number.
+
+      TBD: REMOVE THIS KEY
```
byendan, line 112: Should this be removed in this ticket? I was trying to follow the conversation with Carson but wasn't clear if we were leaving this for the next update.

byendan (commented):
Looks great, but I noticed a few TBDs that I left comments on.

source/api_documentation/network_integration/js_tags/js_revision_parameters.rst
SomeSayKosm: I think this is actually just the internal ID based on a query utilizing it: https://github.com/Invoca/web/blob/18481a657463a8ebf154a9b8c7c36db5cc73af5a/app/models/js_tag/tag.rb#L387

source/api_documentation/network_integration/js_tags/js_revision_parameters.rst
SomeSayKosm: In my reply to the question regarding the `id_from_network` versus the internal `id`, I was thinking that of the two, we should display the external id, since that appears to be the one we operate on in the [advertiser_campaign API](https://developers.invoca.net/en/2019-05-01/api_documentation/network_integration/advertiser_campaigns/index.html) and thus is probably the more valuable of the two.  

I don't think it the worst thing if we include both but I believe they'll only be interested in knowing the id_from_network based on the above campaign documentation.

byendan (approved):
Carson's feedback makes sense to me, I'm ready to approve this.

SomeSayKosm (approved):

PR #350, opened 2023-11-22
Story 16732 update endpoints to work with tag customer phone number associations

SomeSayKosm:
https://invoca.atlassian.net/browse/STORY-16732
Screenshot 2023-11-22 at 10 47 06 AM

Address the new parameter introduced in https://github.com/Invoca/web/pull/22374.

Additionally, removed the explanation around the include_inactive parameter in the CustomerPhoneNumber API.  This functionality was already removed as a part of https://github.com/Invoca/web/pull/22271 where more robust and general filtering was added.


byendan (approved):
Looks good to me 👍

byendan (approved):
Updates look good 👍

PR #351, opened 2023-11-29
STORY-16732: Add notices for creating Customer Phone Numbers

SomeSayKosm:
https://invoca.atlassian.net/browse/STORY-16732

Added two notices.  
One to the RingPool Index, explaining how to create a Destination/CPN via the POST action.  
image
And another to the Destination endpoint, explaining how Creating/Deleting Destinations can be handled and linking to the RingPool documentation page.
image

### DIFF
--- a/source/api_documentation/network_integration/customer_phone_numbers/index.rst
+++ b/source/api_documentation/network_integration/customer_phone_numbers/index.rst
@@ -121,7 +121,7 @@ Endpoint:
   <br/>
   <div class="alert alert-info">
     <b>NOTE: Creating & Deleting Destinations</b></br>
-    Creating a Destination via this endpoint isn't possible.  One can be made alongside a RingPool that it'll be associated with via the RingPool endpoint (<a href="https://developers.invoca.net/en/2022-08-01/api_documentation/network_integration/ringpools/index.html"><b>Documentation</b></a>).<br/>
+    Creating a Destination via this endpoint isn't possible. A Destination can be made alongside a RingPool that it'll be associated with via the RingPool endpoint (<a href="https://developers.invoca.net/en/latest/api_documentation/network_integration/ringpools/index.html"><b>Documentation</b></a>).<br/>
     <br/>
     The Destination can be deleted via this endpoint, but if the associated RingPool is going to be deleted as well, you can save a request by utilizing the delete_detected_phone_number parameter in the RingPool endpoint.
   </div>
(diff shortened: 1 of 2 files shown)


source/api_documentation/network_integration/customer_phone_numbers/index.rst
```diff
@@ -114,3 +114,15 @@ Endpoint:
...
    :description: Delete a Destination
    :page: delete_customer_phone_number
+
+.. raw:: html
+
+  <br/>
+  <br/>
+  <div class="alert alert-info">
+    <b>NOTE: Creating & Deleting Destinations</b></br>
+    Creating a Destination via this endpoint isn't possible.  One can be made alongside a RingPool that it'll be associated with via the RingPool endpoint (<a href="https://developers.invoca.net/en/2022-08-01/api_documentation/network_integration/ringpools/index.html"><b>Documentation</b></a>).<br/>
```
omeedrj: 🔧 `One can be made alongside a RingPool that it'll be associated with via the RingPool endpoint` -> `A Destination can be made alongside a RingPool that it'll be associated with via the RingPool endpoint`

omeedrj: ♻️ Instead of referencing `2022-08-01` directly, can we reference `latest` so that these docs don't need to be kept up-to-date as new API versions are created in the future?

Alternatively if we don't want to use `latest`, could we mimic this pattern? https://github.com/Invoca/developer-docs/blob/8ad02e7479c8c78277fa0a088885621a374445e7/source/doc_versions.py

source/api_documentation/network_integration/ringpools/index.rst
```diff
@@ -94,6 +94,18 @@ Endpoint:
    :page: delete_ring_pool
 
 
+.. raw:: html
+
+  <br/>
+  <br/>
+  <div class="alert alert-info">
+    <b>NOTE: Creating Destinations</b></br>
+    If a valid phone number is given for destination_phone_number is given as part of the POST request, a Destination, aka Customer Phone Number, will be generated and associated with the created RingPool.<br/>
```
omeedrj: 🔧 `If a valid phone number is given for destination_phone_number is given as part of the POST request` -> `If a valid phone number is given for destination_phone_number as part of the POST request`

byendan (approved):
This looks right to me, makes sense why it's not showing up now.

omeedrj (changes requested):
(Reviewed with [CREG](https://github.com/erikthedeveloper/code-review-emoji-guide))

omeedrj (approved):

PR #376, opened 2024-05-20
STORY-11379: remove crapi from the docs

jebentier:
This is removing the deprecated (and now deleted) CRAPI endpoints from the API documentation. This is just getting it removed from the latest API version, but I'll cherry pick the change to older versions of the docs as well.

## Checklist

- [x] Test the documentation changes on readthedocs as a private branch
- [x] If changing general content, have agreement on whether to apply to latest version or all versions (if all versions, provide links to the related PRs below)


nburwell (approved):
Nice, thanks! 

> This is just getting it removed from the latest API version, but I'll cherry pick the change to older versions of the docs as well.

For the older versions, I'm on the fence between a straight cherry pick of these deletions, or if we leave the page with the banner that directs to the Signal API and then remove all content under it. That way old bookmarked URLs would still route, and users would get the info of where to go now, versus just getting a 404. 

It's minor though for this API, given it's very limited usage in the past several years.

jebentier:
Yea, given that on older versions we didn't even add the `deprecated`, I'm on the fence about delete vs redirect. Delete is definitely the easier, but I don't know which is better in terms of customer experience.

